PR #137, opened 2021-09-17
CI/CD: Add Basic Unit tests via Azure pipeline

evberrypi:
Adding a .azure directory with one pipeline that runs formatting checks and unit tests. 
I will open an issue for the still missing creation of a release.

### DIFF
--- a/.azure/azure-build-pipeline.yml
+++ b/.azure/azure-build-pipeline.yml
@@ -14,12 +14,6 @@ stages:
         matrix:
           Python3.6:
             python.version: '3.6'
-          Python3.7:
-            python.version: '3.7'
-          Python3.8:
-            python.version: '3.8'
-          Python3.9:
-            python.version: '3.9'
         maxParallel: 3
       steps:
       - task: UsePythonVersion@0
@@ -40,5 +34,5 @@ stages:
         displayName: 'Check formatting'
       - script: |   
           source .venv/bin/activate 
-          ./run test
+          ./run test all unit
         displayName: 'Run Unit tests'
--- a/.azure/azure-build-pipeline.yml
+++ b/.azure/azure-build-pipeline.yml
@@ -10,34 +10,33 @@ stages:
     - job: RunTests
       pool:
         vmImage: ubuntu-20.04
-        strategy:
-          matrix:
-            Python3.7:
-              python.version: '3.7'
-            Python3.8:
-              python.version: '3.8'
-            Python3.9:
-              python.version: '3.9'
-          maxParallel: 3
+      strategy:
+        matrix:
+          Python3.7:
+            python.version: '3.7'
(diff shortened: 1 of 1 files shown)


.azure/azure-build-pipeline.yml
```diff
@@ -0,0 +1,50 @@
...
+          ./run test
+        displayName: 'Run Unit tests'
+      - script: |
+          python -m pip install -r docs/requirements.txt
+          export PYTHONPATH=$(pwd)
+          cd docs/
+          make clean
+          make html 2>&1 | tee tmp_log
+          if [[ $(grep -c "WARNING:" tmp_log) != 0 ]]; then echo "found warnings"; grep "WARNING:" tmp_log; exit 1; fi
+        displayName: 'Build Docs'
```
gigony, line 50: need a blank line after this.

.azure/azure-build-pipeline.yml
```diff
@@ -0,0 +1,50 @@
...
+        displayName: 'Setup development environment'
+      - script: | 
+          source .venv/bin/activate 
+          ./run check -f
+        displayName: 'Check formatting'
+      - script: |   
+          source .venv/bin/activate 
+          ./run test
+        displayName: 'Run Unit tests'
+      - script: |
```
gigony, line 43: Please remove building Docs. we don't need this.

.azure/azure-build-pipeline.yml
```diff
@@ -0,0 +1,50 @@
...
+  - stage: 'Test'
+    jobs:
+    - job: RunTests
+      pool:
+        vmImage: ubuntu-20.04
+      strategy:
+        matrix:
+          # Python3.7:
+          #   python.version: '3.7'
+          Python3.8:
```
gigony, line 17: Could we change it to 3.6?
our minimum supported Python version is 3.6 and we want to verify it on Python 3.6
(people may use a Python syntax that is not available at Python 3.6).

evberrypi: Yes, just added python3.6

.azure/azure-build-pipeline.yml
```diff
@@ -0,0 +1,44 @@
...
+          source .venv/bin/activate 
+          ./run setup
+        displayName: 'Setup development environment'
+      - script: | 
+          source .venv/bin/activate 
+          ./run check -f
+        displayName: 'Check formatting'
+      - script: |   
+          source .venv/bin/activate 
+          ./run test
```
gigony, line 43: tests include system test which takes long time.
Could we execute only unit tests?
```
./run test all unit  # execute unit tests only
```

evberrypi: Done, will test everything in weekly build pipeline

PR #140, opened 2021-09-18
Segmentation mask image should have uint8 type, for both the saved nii.gz file and in memory Image 

MMelQin:
It is known that the MONAI transform class, `SaveImaged`, uses `np.float32` as the output image data type if none specified.

For segmentation image, the mask image invariantly has `uint8 `for each segment. Furthermore, it has become known that external render programs, e.g. Clara Render Server, assumes `uint8 `segmentation mask image and would fail with other data type.

In addition, the App SDK DICOM Segmentation Writer consumes the in memory segmentation image, and even though `float32` type is handled, it is better to save the image with uint8 for reducing mem usage as well as being consistent.


gigony (approved):
looks good to me!

PR #143, opened 2021-09-21
Change the name of 'input' and 'output' parameters in Operator.compute() method

### LINKED ISSUE #131: [BUG] `input` is a Python built-in. Should it be avoided as a function name?
**Describe the bug**

`input` is a [Python built-in](https://docs.python.org/3/library/functions.html#input). The app SDK has a method called `monai.deploy.core.input` (used as a decorator for `Operator`s). In my opinion, this is generally a bad idea, since importing it clobbers the built-in function name and could cause confusion and weird bugs in user code. Also, but less importantly, many syntax highlighters highlight built-ins differently from other functions/variables.

**Steps/Code to reproduce bug**
```python
something = input()  # built-in

from monai.deploy.core import input  # some editors would highlight this differently, I guess github is smarter

something_else = input()  # behaves differently now due to clobber!
```

**Expected behavior**
As a general rule, I would always avoid clobbering built-ins with identical function/class names in a public API. I would suggest renaming this function/decorator. Maybe `app_input` instead? Whatever is chosen, `output` should probably be changed accordingly too, even though it is not the name of a built-in.

**Environment details (please complete the following information)**

- OS/Platform: Any
- Python Version: Any
- Method of MONAI Deploy App SDK install: Any
- SDK Version: 0.1.0a2+33.gfb2354d

**Additional context**

gigony:
- Update docs and code for compute() method of Operator class
  - 'input' -> 'op_input'
  - 'output' -> 'op_output'
- Use fully qualified method name for decorators
  - `import monai.deploy.core as md`
  - `@md.input(...)`
  - `@md.output(...)`

Resolves #131.


MMelQin (approved):
Looks good.

PR #148, opened 2021-09-23
Udpate the spleen seg application notebook

MMelQin:
The Spleen Seg Application notebook has been updated and tested running locally.

Note that it assume the DICOM files are in `dcm` folder, and the TS model at `model/spleen.ts`. May need to come up with the download step once we are clear about where from (the shared Google Drive?)


gigony (approved):
Thanks Ming!

Let's merge this. 
I see some typos but can handle them on my PR.

Issue #165, opened 2021-10-08
[BUG] ERROR: Failed to fetch MAP manifest.


jpcenteno80:
**Describe the bug**
Above error displayed when trying to run packaged app.

**Steps/Code to reproduce bug**
```
# Inside monai-deploy-app-sdk/examples/apps/
monai-deploy package ./simple_imaging_app -t simple_app:latest

# output:
Building MONAI Application Package... Done
[2021-10-08 11:04:07,763] [INFO] (app_packager) - Successfully built simple_app:latest

monai-deploy run simple_app:latest input output # where input has brain_mr_input.jpg and output is an empty directory

# output:
Checking dependencies...
--> Verifying if "docker" is installed...

--> Verifying if "simple_app:latest" is available...

Checking for MAP "simple_app:latest" locally
"simple_app:latest" found.

Reading MONAI App Package manifest...
 > export '/var/run/monai/export/' detected
fatal: Access to the path '/var/run/monai/export/config/pkg.json' is denied.
ERROR: Failed to fetch MAP manifest.
Execution Aborted
```

**Expected behavior**
App should run.

**Environment details (please complete the following information)**

- OS/Platform: Ubuntu 18
- Python Version: 3.9.7
- Method of MONAI Deploy App SDK install: `pip install monai-deploy-app-sdk`
- SDK Version: 0.1.0

**Additional context**
This is an AWS instance where I created my own user. App was packaged under my user.

gigony:
Thanks @jpcenteno80  for the feedback!

`monai-deploy-app-sdk` package and its dependent pip packages are copied/added by default when `monai-deploy package` command is used.

https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/41eddc4bc4a7a9e9839377622da75334d323a2b3/monai/deploy/packager/util.py#L155-L158

https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/41eddc4bc4a7a9e9839377622da75334d323a2b3/monai/deploy/packager/util.py#L165-L167

https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/20bb941d67d9e12d64d3f99dd9001d15d536ae70/monai/deploy/packager/templates.py#L47

`monai-deploy-app-sdk` is a general-purpose deploy framework on the healthcare domain (such as genomics/biochemistry workflows) so it doesn't depend on `monai` package (MONAI Core) and it needs to be manually specified if your application uses `monai` package.

[...]

PR #168, opened 2021-10-09
Build package with local UID/GID and fast-retrieve app/package information

gigony:
MAP is created with the default UID/GID (1000) by default.
However, if the user has UID which is not 1000 such as 1001, the local input/output folder is mounted with the host's input folder permission (1001) by App Runner and causes the following message when MAP is executed.

```bash
...
Reading MONAI App Package manifest...
 > export '/var/run/monai/export/' detected
fatal: Access to the path '/var/run/monai/export/config/pkg.json' is denied.
ERROR: Failed to fetch MAP manifest.
Execution Aborted
```

This patch builds a Docker image with local UID/GID to resolve the issue.

Also, it retrieves app/package information with the faster method if the OS is not Windows.
(copying app.json/pkg.json through `docker cp` command is faster than launching container to download config file).

```python
        if sys.platform == "win32":
            cmd = f"docker run --rm -a STDOUT -a STDERR -v {info_dir}:/var/run/monai/export/config {map_name}"
        else:
            cmd = f"""docker_id=$(docker create {map_name})
docker cp $docker_id:/etc/monai/app.json {info_dir}/app.json
docker cp $docker_id:/etc/monai/pkg.json {info_dir}/pkg.json
docker rm -v $docker_id > /dev/null
"""
```
The bug is confirmed and tested by creating AWS EC2 instance:

```
AMI: ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20210430
Instance type: t2.large
Storage: 20 GB 
```
Due to large base image size, only 1.3GB was left after exercise the Tutorial 1 (examples/apps/simple_imaging_app):
```
(monai) gbae@ip-172-31-0-144:~/monai-deploy-app-sdk$ df
Filesystem     1K-blocks     Used Available Use% Mounted on
/dev/root       20263484 18888788   1358312  94% /
devtmpfs         4069584        0   4069584   0% /dev
tmpfs            4075372        0   4075372   0% /dev/shm
tmpfs             815076      832    814244   1% /run
tmpfs               5120        0      5120   0% /run/lock
tmpfs            4075372        0   4075372   0% /sys/fs/cgroup
/dev/loop0         56832    56832         0 100% /snap/core18/1997
/dev/loop1         34176    34176         0 100% /snap/amazon-ssm-agent/3552
/dev/loop2         33152    33152         0 100% /snap/snapd/11588
/dev/loop3         72192    72192         0 100% /snap/lxd/19647
tmpfs             815072        0    815072   0% /run/user/1001
```

This patch is verified in the AWS EC2 instance by using the following test package:

```bash
pip install -i https://test.pypi.org/simple/ monai-deploy-app-sdk==0.0.165
```

Resolves https://github.com/Project-MONAI/monai-deploy-app-sdk/issues/165

P.S. It might be better to use 'docker' client library instead of executing docker with shell command. We will address the issue later.

### DIFF
--- a/monai/deploy/packager/util.py
+++ b/monai/deploy/packager/util.py
@@ -229,10 +229,12 @@ def build_image(args: dict, temp_dir: str):
         dockerignore_file.write(docker_ignore_template)
 
     # Build dockerfile into an MAP image
-    docker_build_cmd = ["docker", "build", "-f", docker_file_path, "-t", tag, temp_dir]
+    docker_build_cmd = f'''docker build -f "{docker_file_path}" -t {tag} "{temp_dir}"'''
+    if sys.platform != "win32":
+        docker_build_cmd += """ --build-arg MONAI_UID=$(id -u) --build-arg MONAI_GID=$(id -g)"""
     if no_cache:
-        docker_build_cmd.append("--no-cache")
-    proc = subprocess.Popen(docker_build_cmd, stdout=subprocess.PIPE)
+        docker_build_cmd += " --no-cache"
+    proc = subprocess.Popen(docker_build_cmd, stdout=subprocess.PIPE, shell=True)
 
     spinner = ProgressSpinner("Building MONAI Application Package... ")
     spinner.start()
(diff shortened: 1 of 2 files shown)


MMelQin:
Yep, I requested Docker SDK be used at the start of the project.

PR #169, opened 2021-10-09
Make InMemImageReader work even when monai is not used/available

gigony:
If the user's application doesn't depend on `monai` and monai is not installed, it causes an error when the following statement is executed by the interpreter.

https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/eebb284d3130e122f933f5241509d0a438699d9d/monai/deploy/operators/monai_seg_inference_operator.py#L239

This patch sets `ImageReader` to `object` if `monai` library is not available so the user's app doesn't require `monai` app if not using.

### DIFF
--- a/monai/deploy/operators/monai_seg_inference_operator.py
+++ b/monai/deploy/operators/monai_seg_inference_operator.py
@@ -20,9 +20,11 @@ torch, _ = optional_import("torch", "1.5")
 np_str_obj_array_pattern, _ = optional_import("torch.utils.data._utils.collate", name="np_str_obj_array_pattern")
 Dataset, _ = optional_import("monai.data", name="Dataset")
 DataLoader, _ = optional_import("monai.data", name="DataLoader")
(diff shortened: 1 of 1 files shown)


gigony:
> The InMemImageReader is a derived class of MONAI ImageReader, hence dependency on MONAI be present.
> 
> 
> 
> This PR is wrong trying to break this dependency at runtime, and should be rejected.

@MMelQin , existing code caused an error even when user's application doesn't use monai.

I faced the issue somehow and 
I think you can reproduce it with simple_imaging_app in a fresh python virtual environment.

Let me also double check it in case the PR cause the side effect you described.

Thank you!

MMelQin:
The example you mentioned should not be loading the monai_seg_inference_operator classes at all, where the InMemImageReader is. So puzzled by the error you saw, while setting the reader to `object` in the seg operator is not the way to address the issue, even if it can help to avoid the symptom (optional_import was already used in the code).

MMelQin (approved):

PR #196, opened 2021-11-01
Remove Azure pipeline, add Github Pipeline + Codecov

evberrypi:
Removes Azure Pipeline in favor of GitHub pipeline (since pipeline is CPU only)
Enables Codecov extension


.github/workflows/pr.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Run Unit tests
+        run: |
+          source .venv/bin/activate
+          ./run test all unit
+      - name: Coverage
+        run: |
+          source .venv/bin/activate
+          coverage xml
+      - name: Upload coverage
+        uses: codecov/codecov-action@v1
```
gigony, line 40: Please update this to v2 as we need to do anyway early next year when v1 is no longer available.

https://github.com/codecov/codecov-action

.github/workflows/pr.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          fail_ci_if_error: false
+          file: ./coverage.xml
```
gigony, line 43: nit: no ends with blank link

gigony (approved):
Looks good to me. Thanks Everett!
Approved with some comments.

.github/workflows/pr.yml
evberrypi: Updated

PR #199, opened 2021-11-10
Remove MONAI user ownership on MAP folders

KavinKrishnan:
Removing fixed `monai` user/privilege's on application package image

### DIFF
--- a/monai/deploy/packager/templates.py
+++ b/monai/deploy/packager/templates.py
@@ -12,47 +12,36 @@
 COMMON_FOOTPRINT = """
     USER root
 
-    RUN pip install --no-cache-dir --upgrade setuptools==57.4.0 pip==21.2.4 wheel==0.37.0
+    RUN pip install --no-cache-dir --upgrade setuptools==57.4.0 pip==21.2.4 wheel==0.37.0 monai-deploy-app-sdk==0.2.0
 
-    RUN groupadd -g $MONAI_GID -o -r monai
-    RUN useradd -g $MONAI_GID -u $MONAI_UID -m -o -r monai
-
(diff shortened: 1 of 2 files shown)


slbryson (approved):
Agreed on the approach.

gigony:
@KavinKrishnan
Could you please sign this commit and update the description of this PR saying the context of this PR?

I wonder to what extent does this PR solves the issue described in https://github.com/Project-MONAI/monai-deploy-app-sdk/issues/202?

KavinKrishnan:
> @KavinKrishnan Could you please sign this commit and update the description of this PR saying the context of this PR?
> 
> I wonder to what extent does this PR solves the issue described in #202?
> 
> Do we just want to run the container as root with this PR?

Yes, and think it is ok. It seems to be a bigger issue having the packager fix permissions in the image around one user `monai`. 

In a future MR, we could provide a way for the user to decide the user when packaging the image. However in the meantime, allowing them to run the container should suffice.

I think the alternative you suggested with a told `gosu` and making each user a super user may not be the best practice.

MMelQin:
Generally for application security, especially FS access controls, use a group to grant permissions to the resources, and on deployment and at runtime, a specific service account just needs to be in this defined group. The application package does not need to pin down the specific user, which is a deployment/runtime concern.

Application understands the required permission or ACL the application specific resources (Docker image files etc, and the mapped folder on host), what specific group, and service account, to use is more the deployment platform configuration concern, to ensure the service account has the required permissions to the resources, without being granted elevated privilege.

whoisj:
LGTM :shipit:

whoisj (approved):

MMelQin:
Merge this PR for now, with the understanding of having #202 to address the account and permission issue.

PR #208, opened 2021-11-24
Fixed minor dependencies for fresh deployment

vikashg:
Here are some changes and dependencies that needs to be added to make the 03_segmentation_app and 04_mis_tutorial notebooks work.

PR #211, opened 2021-11-26
Bump numpy >= 1.20 to avoid MKL error in conda env, and add --upgrade for pip install app sdk

MMelQin:
It was found out that when running Jupyter-lab within a environment set up using conda, pytorch encountered error related to MKL, see [pytorch issue](https://github.com/pytorch/pytorch/issues/37377).

A previous [PR ](https://github.com/Project-MONAI/monai-deploy-app-sdk/pull/208) modified the Jupyter notebook adding env var ['MKL_THREADING_LAYER'] = 'GNU' to work around the MKL issue, which is one of the suggested fixes.

A cleaner fix, as also [stated in the said issue](https://github.com/pytorch/pytorch/issues/37377#issuecomment-807271090), is to upgrade to numpy 1.20 to avoid the MKL issue without needing the env var workaround.

The example segmentation examples had already been tested with numpy >= 1.20, so it is the safe and correct fix to bump the numpy version >= 1.20, in the apps, as well as in the Jupyter notebooks that copied the app code.

Also added `--upgrade` option for pip install monai-deploy-app-sdk, in case the env had an earlier version of the sdk installed.

### DIFF
--- a/examples/apps/ai_livertumor_seg_app/livertumor_seg_operator.py
+++ b/examples/apps/ai_livertumor_seg_app/livertumor_seg_operator.py
@@ -34,7 +34,7 @@ from monai.transforms import (
 @md.input("image", Image, IOType.IN_MEMORY)
 @md.output("seg_image", Image, IOType.IN_MEMORY)
 @md.output("saved_images_folder", DataPath, IOType.DISK)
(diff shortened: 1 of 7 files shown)


gigony (approved):
Thanks Ming! Looks good to me!

PR #222, opened 2021-12-04
Change the model name in the Spleen seg example dataset to model.ts

MMelQin:
This is to avoid confusion as the .pt and .ts are used by some model sources to distinguish PyTorch and TorchScript model files, e.g. Clara MMAR.

A new dataset zip file has been uploaded to Google drive, and the Jupyter notebooks in the tutorials have been updated to download the new zip file and use the updated model file name. The original dataset zip file is still available on the Google drive, in case existing version of the Spleen Seg Jupyter notebook is used.

Noted again that Jupyter notebooks 04 and 05 copied some steps from the Spleen Seg; the duplication inevitably causes duplicated effort when the source changes.      

### DIFF
--- a/docs/source/getting_started/tutorials/03_segmentation_app.md
+++ b/docs/source/getting_started/tutorials/03_segmentation_app.md
@@ -50,25 +50,25 @@ cd monai-deploy-app-sdk
 # Install monai-deploy-app-sdk package
 pip install monai-deploy-app-sdk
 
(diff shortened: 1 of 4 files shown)


vikashg (approved):
I ran everything on my desktop and the changed notebook worked as expected.

dbericat:
@MMelQin do you mind updating the comment with links to both new .zip and old in the Google shared folder? Easier to follow the PR. 

Also please, add labels and move to 0.3? Thanks!

MMelQin:
@dbericat Please see file changes in this PR; the zip link is encoded, adding them in the comment does not help much at all. The diff shows exactly what have changed.

dbericat:
> 

My bad @MMelQin. You are right. Thanks!

Issue #171, opened 2021-10-09
The App SDK domain specific Image class needs to be improved as it lacks consistency of the image array it serves out


MMelQin:
**Is your feature request related to a problem? Please describe.**
There is a fundamental issue with the Image class as the it lacks precise semantics of the dims of its encapsulated pixel image array, i.e. for a 3D image, which DIM is meant to be the axial/slice direction in DICOM speaking, first DIM to represent the instance, or the last DIM? What happens if some code creates the Image with a pixel array that has the instance as the first DIM.

Anyway, we know the issue with this first version of Image. I had to pay close attention to the what the original image type is, and how and what image array is used when instantiating Image, e.g. DICOM to Volume get a image pixel array with depth dim (slice wise) first, but I had to change that code to swap the depth to last. There is also the swapping of width and height, to as to ensure the shape out of this image.asnumpy() would the same as those from ITK or Nibabel.

This PR may resolve the symptom somewhat, but it is at the mercy of the encapsulated image pixel array. I will ensure an issue is created to address the Image class; I don't think it is usable as intended, as it now simply provides an in-memory storage to the pertinent image array. 

**Describe the solution you'd like**
Need to tighten the instantiation of the Image class
Provide consistency over access method to get the encapsulated image array as Numpy array, similar to ITK and NiBabel

**Describe alternatives you've considered**
Well, treating the Image simply as an in-mem storage, meaning the creator and consumer of the Image object know exactly the shape of the encapsulated image array. 
**Additional context**
Built-in operator, [PNGConverterOperator](https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/main/monai/deploy/operators/png_converter_operator.py) has already fall victim of Image, as the operator code assumes the in-mem Image object as the depth as the first dim, however, when used with other operators, image.asnumpy() simply serves out the array that was used to instantiate, and has the depth as the 3rd dim.

PR #235, opened 2022-01-11
Add Clara Viz operator.

AndreasHeumann:
Signed-off-by: Andreas Heumann <aheumann@nvidia.com>


vikashg:
Hey @AndreasHeumann, I tried to read a simple MR image (MNI template) using the monai transform and I wanted to display it on the [jupyter notebooks](https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/vikash/testclaraviz/notebooks/tutorials/06_integrating_clara_viz.ipynb) but nothing happens. I followed the documentation [here](https://docs.nvidia.com/clara-viz/quickstart.html#render-ct-data-from-python).

AndreasHeumann:
Hi @vikashg. I tried the jupyter notebook you provided and indeed the Python kernel crashes for me as well. But it works if i remove the line `print(Renderer(image))`. I'm not sure why it crashes when doing the print in that way but `print(renderer)` for example works. I'll open an issue in Clara Viz on this.
image

vikashg:
Hi @AndreasHeumann, So I removed the line `print(renderer)` line but still there is no display on my notebook. Can you share maybe your environment details. I wish I could give more info other than the terminal output 
```
Opening in existing browser session.
[651765:651765:0100/000000.850456:ERROR:angle_platform_impl.cc(44)] Display.cpp:894 (initialize): ANGLE Display::initialize error 12289: Unsupported GLX version (requires at least 1.3).
[651765:651765:0100/000000.850557:ERROR:gl_surface_egl.cc(782)] EGL Driver message (Critical) eglInitialize: Unsupported GLX version (requires at least 1.3).
[651765:651765:0100/000000.850576:ERROR:gl_surface_egl.cc(1382)] eglInitialize OpenGL failed with error EGL_NOT_INITIALIZED, trying next display type
[651765:651765:0100/000000.856389:ERROR:angle_platform_impl.cc(44)] Display.cpp:894 (initialize): ANGLE Display::initialize error 12289: Unsupported GLX version (requires at least 1.3).
[651765:651765:0100/000000.856459:ERROR:gl_surface_egl.cc(782)] EGL Driver message (Critical) eglInitialize: Unsupported GLX version (requires at least 1.3).
[651765:651765:0100/000000.856484:ERROR:gl_surface_egl.cc(1382)] eglInitialize OpenGLES failed with error EGL_NOT_INITIALIZED
[651765:651765:0100/000000.856504:ERROR:gl_ozone_egl.cc(20)] GLSurfaceEGL::InitializeOneOff failed.
[W 19:12:37.813 NotebookApp] Notebook 06_integrating_clara_viz.ipynb is not trusted
[W 19:12:37.841 NotebookApp] 404 GET /nbextensions/clara-viz-widgets/extension.js?v=20220114191109 (127.0.0.1) 20.650000ms referer=http://localhost:8888/notebooks/06_integrating_clara_viz.ipynb
[I 19:12:38.067 NotebookApp] Kernel started: ee55c289-ab1f-42ab-ac14-fb5b2d44256f, name: python3
[W 19:12:38.077 NotebookApp] 404 GET /nbextensions/widgets/notebook/js/extension.js?v=20220114191109 (127.0.0.1) 2.080000ms referer=http://localhost:8888/notebooks/06_integrating_clara_viz.ipynb
[W 19:13:21.845 NotebookApp] 404 GET /static/clara-viz-widgets.js?v=20220114191109 (127.0.0.1) 6.400000ms referer=http://localhost:8888/notebooks/06_integrating_clara_viz.ipynb
[I 19:14:38.734 NotebookApp] Saving file at /06_integrating_clara_viz.ipynb

```

AndreasHeumann:
The first few lines come from the browser not being able to initialize OpenGL. Did you run the Jupyter Lab command from a ssh connection? Usually I run it on ssh with `jupyter lab --no-browser --ip=` on my LInux desktop and the open the browser on my Windows Laptop and navigate to the IP address of the desktop, e.g. `http://192.168.178.44:8888/`.
On my desktop I'm using Ubuntu 20.04, I'm using a conda environment with Monai Deploy App SDK, Jupyter Lab and Clara Viz installed.

vikashg:
Hi @AndreasHeumann, So yes it works with `jupyter-lab` but not with `jupyter-notebook`

AndreasHeumann:
jupyter-lab and jupyter-notebook are using different interfaces for the widgets. Clara Viz should support both, but we had issues with that in the past. I'll check why it does not work with jupyter-notebook.

AndreasHeumann:
I filed https://github.com/NVIDIA/clara-viz/issues/9 for the bug that the Clara Viz widget is not working in Jupyter Notebook (only in Lab).

AndreasHeumann:
@vikashg I just released Clara Viz 0.1.2. That version fixes the crash you were seeing with `print(Renderer(data))` and now the widget also works in Jupyter Notebook (not just Jupyter Lab).

vikashg:
Thanks @AndreasHeumann I will check it out today/tomorrow.

monai/deploy/operators/clara_viz_operator.py
```diff
@@ -0,0 +1,133 @@
...
+        if not input_image:
+            raise ValueError("Input image is not found.")
+        input_seg_image = op_input.get("seg_image")
+        if not input_seg_image:
+            raise ValueError("Input segmentation image is not found.")
+
+        # build the data definition
+        data_definition = DataDefinition()
+
+        data_definition.arrays.append(self._build_array(input_image, "DXYZ"))
```
MMelQin, line 98: I've merge this PR, #238 , so now in-memory input image array asnumpy() now has index ordering of DHW, and the seg image also has this ordering. The numpy array of images coming from DICOM loader, and leaving the segmentation operator are now supposed to be similar to the numpy array from ITK.

AndreasHeumann: Thanks, @MMelQin. I updated the Pull Request and removed the workaround.

monai/deploy/operators/clara_viz_operator.py
```diff
@@ -0,0 +1,133 @@
...
+    This operator uses Clara Viz to provide interactive view of a 3D volume including segmentation mask.
+    """
+
+    def __init__(self):
+        """Constructor of the operator."""
+        super().__init__()
+
+    @staticmethod
+    def _build_array(image, order):
+        numpy_array = image.asnumpy().copy()
```
AndreasHeumann, line 41: @MMelQin @gigony I have a strange bug here. If I don't copy the segmentation numpy array data here then the data is wrong. Clara Viz copies the data when the Widget set_datafinition() function is called below in line 101.
Not sure how this could happen. Python is ref-counting objects, it should not be possible that the data is overwritten between here and line 101.

MMelQin: @AndreasHeumann you have actually brought a very important point, in general with regard to the App SDK Image object state changes.

In your code, does the DataDefinition.Array simply refer to the passed in numpy_array, and modify it within  DataDefinition.Array.flip_axes? If yes, then I'd think the original ndarray would be modified.

On to the App SDK [Image](https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/main/monai/deploy/core/domain/image.py) class itself, which at present is a very primitive wrapper/accessor. Its asnumpy() function simply returns the _data object, which is ArrayLike and set via the Image constructor. Modifying the object returned asnumpy() would directly modify the encapsulated object.

In the inference application, the DICOMSeriesToVolumeOperator create the (density) Image object (type int16), the Inference operator's InMemImageReader get the object via asnumpy() and exposes it via its get_data() without copying, and the image pre-processing before inference directly transforms the data from calling get_data(), so the (density) image may get modified. It is not an issue (yet) for the inference app as the original (density) image is not needed once inference is done, but may cause issues for the ClaraViz operator instances, which may try to get to the input (density) image at different stage of the app execution.

We can document this behavior and require the accessor of Image.asnumpy() not to modify the return, or make a copy to modify, but the best would be to make the function(s) on the Image class more explicit.

AndreasHeumann: Clara Viz is not modifying the data, it's just copying it. Density is fine, it's just the segmentation data which is wrong. Maybe when doing a copy() the resulting memory is continuous, internally Clara Viz is always using continuous memory and I have a copy function from non-continuous Python data to the continuous internal memory. Maybe there is a bug in that code.

Unfortunately there is no concept of immutable or constant objects in Python, so documenting the accessor is the only thing we can do.

MMelQin: NumPy copy is not a deep copy, maybe `copy.deepcopy` will work better if needed?

AndreasHeumann: You are right. But using the copy() already makes it go down the continuous path. And I just found the bug in the non-continuous copy function, everything now works without making a copy of the array.

I also found this https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html?highlight=writeable. We can make numpy array data immutable, just tried it out and it works:

```
>>> a = numpy.zeros(100)
>>> a[0] = 12
>>> a.flags.writeable = False
>>> a[0] = 12
  File "<stdin>", line 1, in <module>
ValueError: assignment destination is read-only
>>> a.flags.writeable = True
>>> a[0] = 12
>>>
```

MMelQin (approved):
Look good to me, and I tested out this operator in the Spleen Seg App.

monai/deploy/operators/clara_viz_operator.py
MMelQin: @AndreasHeumann There is already an open issue https://github.com/Project-MONAI/monai-deploy-app-sdk/issues/171 to enhance the Image class. Properly managing the state of its encapsulated data object (in most cases numpy array) should be part of it. Thanks.

dbericat:
Is it fair to say we already have a good Clara Viz operator? @MMelQin what are the plans to update the Jupyter notebooks with it?

MMelQin:
In the upcoming v0.3 release, yes.

vikashg:
We have  to prepare a jupyter notebook for tutorials and maybe update the demos 4 and 5 with some clara-viz operators. to fully integrate it.

Issue #187, opened 2021-10-19
DICOMSeriesToVolumeOperator fails when DICOM instances do not have (0028, 1052), Rescale Intercept attribute


MMelQin:
**Describe the bug**
For an example DICOM CT series (likely converted into DICOM from NIfTI), where there is no tag (0028, 1052) Rescale Intercept, the DICOMSeriesToVolumeOperator fails as it does not handle the absence of this attribute, as seen from the log.

```
  File "/usr/local/lib/python3.6/dist-packages/monai/deploy/core/executors/single_process_executor.py", line 125, in run
    op.compute(op_exec_context.input_context, op_exec_context.output_context, op_exec_context)
  File "/usr/local/lib/python3.6/dist-packages/monai/deploy/operators/dicom_series_to_volume_operator.py", line 37, in compute
    voxel_data = self.generate_voxel_data(dicom_series)
  File "/usr/local/lib/python3.6/dist-packages/monai/deploy/operators/dicom_series_to_volume_operator.py", line 55, in generate_voxel_data
    intercept = slices[0][0x0028, 0x1052].value
  File "/usr/local/lib/python3.6/dist-packages/monai/deploy/core/domain/dicom_sop_instance.py", line 43, in __getitem__
    return self._sop.__getitem__(key)
  File "/usr/local/lib/python3.6/dist-packages/pydicom/dataset.py", line 916, in __getitem__
    elem = self._dict[tag]
KeyError: (0028, 1052)
```

**Steps/Code to reproduce bug**
Choose a series that is known without the said attribute

**Expected behavior**
The said operator should be able to handle the absence of attribute, treat it as 0.
This attribute, for CT, in the mandatory CT Image Modules is a mandatory attribute, but somehow the input DICOM series is missing this attribute. With it missing, it is safe to assume it being 0.

**Environment details (please complete the following information)**

- OS/Platform: Ubuntu 18.04
- Python Version: 3.7.5
- Method of MONAI Deploy App SDK install: from source
- SDK Version: v0.1

**Additional context**
See DICOM Standards Part3, or this page, https://dicom.innolitics.com/ciods/ct-image/ct-image/00281052

MMelQin:
Addressed with [PR 187](https://github.com/Project-MONAI/monai-deploy-app-sdk/pull/236)

PR #236, opened 2022-01-12
Handle exception case with missing DICOM Rescale Intercept and Slope attributes

MMelQin:
Per Issue [#187](https://github.com/Project-MONAI/monai-deploy-app-sdk/issues/187), added exception handling when parsing Rescale Intercept (0028,1052) and Rescale Slope (0028,1053).

- Rescale Intercept (0028,1052) is of Type 1C. It is required if Modality LUT Sequence (0028,3000) is not present, and shall not be present otherwise.
- Rescale Slope (0028,1053) is also Type 1C, required if Rescale Intercept is present. 
- Rescale Type (0028, 1054) is not parsed or used in the implementation as of now. This attribute is required if Rescale Intercept is present.

These attributes have been observed missing, even in the absence of LUT, in DICOM instances generated with open source utility tools that converts NIfTI image to DICOM. It is safe to use 0 and 1 for the intercept and slope to handle the exception.

The code change has been tested successfully with a DICOM series known to be missing the said attributes.

Note: the build/test failure is due to mypy checking, and the errors are from the runner and packager, not in the modules affected by this PR, as shown below.
```
Checking styles using mypy...
2022-01-12 07:36:06 $ python3 -m mypy --version
mypy 0.931
2022-01-12 07:36:07 $ python3 -m mypy --namespace-packages --explicit-package-bases /home/runner/work/monai-deploy-app-sdk/monai-deploy-app-sdk
monai/deploy/runner/run_command.py:44:5: error: Returning Any from function declared to return "ArgumentParser"  [no-any-return]
monai/deploy/packager/package_command.py:48:5: error: Returning Any from function declared to return "ArgumentParser"  [no-any-return]
monai/deploy/cli/exec_command.py:55:5: error: Returning Any from function declared to return "ArgumentParser"  [no-any-return]
Found 3 errors in 3 files (checked 112 source files)
mypy check failed!
Error: Process completed with exit code 1.
```

### DIFF
--- a/monai/deploy/operators/dicom_series_to_volume_operator.py
+++ b/monai/deploy/operators/dicom_series_to_volume_operator.py
@@ -89,6 +89,7 @@ class DICOMSeriesToVolumeOperator(Operator):
             intercept = slices[0][0x0028, 0x1052].value
         except KeyError:
             intercept = 0
+
         try:
             slope = slices[0][0x0028, 0x1053].value
         except KeyError:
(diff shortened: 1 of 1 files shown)


gigony (approved):
Thank you Ming! Looks good to me!

PR #237, opened 2022-01-12
Fix IPython detection routine

### LINKED ISSUE #206: [BUG] Tutorial Notebook Code Not Working With `inspect`
**Describe the bug**
Code in notebook cells do not have an associated filename, `inspect.getfile` will raise an error if applied to definitions in cells. The notebook `notebooks/tutorials/01_simple_app.ipynb` has this issue where an instance of the class `App` is created without specifying a path. 

**Steps/Code to reproduce bug**
Run the notebook up to the cell below `We can execute the app in the Jupyter notebook.` This cell will raise an error indicating `inspect` couldn't find a filename. If the code of the cell is modified to be `app = App(path='.')` it will function correctly.

**Expected behavior**
The exception from `inspect.getfile` should be caught and a default filename chosen.

**Environment details (please complete the following information)**

- OS/Platform: Ubuntu 20.04
- Python Version: 3.9
- Method of MONAI Deploy App SDK install: pip for deploy sdk, conda for everything else
- SDK Version: 0.2.0

**Additional context**
This may be a Python 3.9 issue, I haven't tested with other versions yet.

gigony:
To detect if the code is run inside Jupyter notebook (IPython) or not, it tried to find whether the class's file name starts with `<ipython-` or not.

Meanwhile the approach worked on Google Colab and some Jupyter Notebook versions. There were some cases where the class's file name is like `/tmp/ipykernel_xxxxx/xxxxxxxxx.py` causing the failure described in #206.

This patch uses a different approach.

If the file path is not available, it tries to see each frame information in the stack to check whether the file name ends with "interactiveshell.py" and the function name is "run_code".

If so, it returns `Path("ipython")` to notify that the class is defined inside IPython.

Fixes #206

### DIFF
--- a/monai/deploy/core/application.py
+++ b/monai/deploy/core/application.py
@@ -95,7 +95,7 @@ class Application(ABC):
             self.path = get_class_file_path(self.__class__)
 
         # Set the runtime environment
-        if str(self.path).startswith("<ipython-"):
+        if str(self.path) == "ipython":
             self.in_ipython = True
         else:
             self.in_ipython = False
--- a/monai/deploy/utils/importutil.py
+++ b/monai/deploy/utils/importutil.py
@@ -99,7 +99,11 @@ def get_application(path: Union[str, Path]) -> Optional["Application"]:
 def get_class_file_path(cls: Type) -> Path:
     """Get the file path of a class.
 
-    If file path is not available, it tries to get the internal class name used in IPython(starting with <ipython-...>).
+    If the file path is not available, it tries to see each frame information
+    in the stack to check whether the file name ends with "interactiveshell.py"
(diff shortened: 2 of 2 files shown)


MMelQin:
This [discussion on StackOverflow](https://stackoverflow.com/questions/23883394/detect-if-python-script-is-run-from-an-ipython-shell-or-run-from-the-command-li) also agrees with your approach.

MMelQin (approved):
Looks good and the approach is also regarded as the most robust per [stack overflow discussion](https://stackoverflow.com/questions/23883394/detect-if-python-script-is-run-from-an-ipython-shell-or-run-from-the-command-li).

PR #238, opened 2022-01-19
Make DICOMSeriesToVolumeOperator consistent with ITK in serving NumPy array

MMelQin:
The noted issue in the generated DICOM Seg instance will be investigated outside of this PR.

The Image object instances in the application pipeline was not completely consistent in its representation of the pixel in NumPy array, especial around the index order, i.e. DWH or DHW. One wonders if the representation is similar to what's provided in other packages, e.g. (Simple) ITK.
  
This PR intends to make the App SDK Image object instances in the processing pipeline of an application more consistent by separating the concerns during image loading of DICOM instance, during inference using MONAI transforms including re-shaping the image array, and the output Image after inference:
1. Consistent with ITK on converting DICOM series instances to volumetric image in DICOMSeriesToVolumeOperator. The NumPy array of the App SDK Image instance has array index order of DHW, and the metadata for pixel spacing and orientation have save values as loaded by (Simple) ITK. This image object is used by downstream operators, e.g. ClaraViz operator, segmentation operator, etc.
2. Consistent with MONAI ITKReader on serving out pixel data and metadata in the InMemImageReader, used in the SegmentationInferenceOperator. MONAI has multiple ImageReader implementations, each for loading certain file formats, and ITKReader is for DICOM instances. To have consistent data representation (with NibabelReader), ITKReader rearranges the NumPy array internally before severing out data, see the [implementation in monai 0.7](https://github.com/Project-MONAI/MONAI/blob/bfa054b9c3064628a21f4c35bbe3132964e91f43/monai/data/image_reader.py#L135). The InMemImageReader is then made consistent with ITKReader.
3. The Image object encapsulating the segmentation inference results is consistent with that of ITK, e.g. the data NumPy array index order is DHW. This Seg image is used by ClaraViz integration operator, the new STL mesh operator, as well as the DICOMSegmentationWriterOperator.

Code changes have been tested:
- The Image loaded from DICOM is as expected.
- The MONAI inference pipeline in the Segmentation Operator works fine with the updated InMemImageReader, and generates the segmentation image in NIfTI same as before.
- _The DICOM Segmentation writer, with the in-mem Image object as input, generates DICOM Seg instance which has multiple segment slices in the same frame. This is an issue, and is still being worked on. Even with the NIfTI image file generated in the inference pipeline, the DICOM Segmentation writer also generates the same (bad) DICOM Seg instance. It is also noted that the  DICOMSegmentationWriterOperator itself did not change._

Update:
The aforementioned DICOMSegmentationWriterOperator issures have been reproduced in App SDK main, tag 0.1.1, and 0.2, so the changes within this PR are not the cause of the errors. It is further noted that the DICOMSegmentationWriterOperator had been tested working properly in release v0.2, except a known issue with a specific input data set. It is unknown what the root cause is as of now, and plan to investigate and fix the root cause outside of this PR.

image

### DIFF
--- a/monai/deploy/operators/monai_seg_inference_operator.py
+++ b/monai/deploy/operators/monai_seg_inference_operator.py
@@ -227,12 +227,13 @@ class MonaiSegInferenceOperator(InferenceOperator):
                     out_ndarray = np.squeeze(out_ndarray, 0)
                     # NOTE: The domain Image object simply contains a Arraylike obj as image as of now.
                     #       When the original DICOM series is converted by the Series to Volume operator,
-                    #       using pydicom pixel_array, the 2D ndarray of each slice is transposed, and the
-                    #       depth/axial direction dim made the last. So once post-transforms have completed,
-                    #       the resultant ndarray for each slice needs to be transposed back, and the depth
-                    #       dim moved back to first, otherwise the seg ndarray would be flipped compared to
-                    #       the DICOM pixel array, causing the DICOM Seg Writer generate flipped seg images.
-                    out_ndarray = out_ndarray.T.astype(np.uint8)  # np.swapaxes(out_ndarray, 2, 0).astype(np.uint8)
+                    #       using pydicom pixel_array, the 2D ndarray of each slice has index order HW, and
+                    #       when all slices are stacked with depth as first axis, DHW. In the pre-transforms,
+                    #       the image gets transposed to WHD and used as such in the inference pipeline.
+                    #       So once post-transforms have completed, and the channel is squeezed out,
+                    #       the resultant ndarray for the prediction image needs to be transposed back, so the
+                    #       array index order is back to DHW, the same order as the in-memory input Image obj.
+                    out_ndarray = out_ndarray.T.astype(np.uint8)
                     print(f"Output Seg image numpy array shaped: {out_ndarray.shape}")
                     print(f"Output Seg image pixel max value: {np.amax(out_ndarray)}")
                     out_image = Image(out_ndarray, input_img_metadata)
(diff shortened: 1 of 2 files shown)


monai/deploy/operators/monai_seg_inference_operator.py
```diff
@@ -290,23 +294,39 @@ def verify_suffix(self, filename: Union[Sequence[str], str]) -> bool:
...
-        A single image is loaded with a single set of metadata as of now."""
+        A single image is loaded with a single set of metadata as of now.
+
+        The App SDK Image asnumpy function is expected to return a NumPy array of index order `DHW`.
+        This is because in the DICOM serie to volume operator pydicom Dataset pixel_array is used to
+        to get per instance pixel NumPy array, with index order of `HW`. When all instances are stacked,
+        along the first axis, the Image NumPy array's index order is `DHW`. ITK array_view_from_image
+        and SimpleITK GetArrayViewFromImage also returns a Numpy array with index order of `DHW`.
+        This NumPy array is then transposed to be consistent with the NumPy array from NibabelReader,
+        which loads NIfTI image and gets NumPy array using the image's get_fdata() function.
```
rahul-imaging, line 313: Can you please mention in the comments what is the final order/format of the pixel data that get_data returns ?

monai/deploy/operators/monai_seg_inference_operator.py
```diff
@@ -290,23 +294,39 @@ def verify_suffix(self, filename: Union[Sequence[str], str]) -> bool:
...
+        # Really does not have anything to do. Simply return the Image object
+        return self.input_image
 
     def get_data(self, input_image):
         """Extracts data array and meta data from loaded image and return them.
 
         This function returns two objects, first is numpy array of image data, second is dict of meta data.
         It constructs `affine`, `original_affine`, and `spatial_shape` and stores them in meta dict.
-        A single image is loaded with a single set of metadata as of now."""
+        A single image is loaded with a single set of metadata as of now.
```
rahul-imaging, line 305: Suggestion for clarifying pixel-order in an Image instance

1. Use an enum to indicate the native pixel-order-format inside an Image instance. We could use an enum to represent all possible image formats:
```
 class PixelFormat(Enum):
 DHW = 1
 WHD = 2
 WDH = 3
```

2.  We can then have a method  inside Image class 

change_pixel_format (destination_pixel_format_type : PixelFormat)

This can be implemented in the next release

rahul-imaging (approved):

monai/deploy/operators/monai_seg_inference_operator.py
MMelQin: Added comments to state that similar to the MONAI ITKReader, the numpy array of the image itself is of index order DHW, but the get_data() impl of the concrete classes will all need to align with MONAI NibabelReader, which return numpy array of the index order WHD(C) when its get_data() is called.

monai/deploy/operators/monai_seg_inference_operator.py
MMelQin: Added comments on the expected/normalized numpy array index order from get_data().

PR #248, opened 2022-01-24
Drop Python 3.6 support

gigony:
As per MONAI Core's [Python 3.6 drop](https://github.com/Project-MONAI/MONAI/pull/3536), drop the support for Python 3.6 from MONAI Deploy App SDK.


MMelQin (approved):
Looks good to me

PR #260, opened 2022-02-02
Add the STL conversion operator

MMelQin:
In this PR
- STLConversionOperator is added which takes in a segmentation Image, generates surface mesh, and saves the image in a STL file.
- A STLConverter class is also added which is used by the aforementioned operator. This class also has a embedded SpatialImage class which wraps the App SDK Image instance and creating ITK image internally.  

An example of the Spleen STL rendering is attached
SpleenSegSTL-2

### DIFF
--- a/monai/deploy/operators/stl_conversion_operator.py
+++ b/monai/deploy/operators/stl_conversion_operator.py
@@ -1,0 +1,445 @@
+# Copyright 2022 MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+import shutil
+import tempfile
+from ast import Bytes
+from pathlib import Path
+from typing import Any
+
+import nibabel as nib
+import numpy as np
+from matplotlib import image
+
+from monai.deploy.utils.importutil import optional_import
+
+sitk, _ = optional_import("SimpleITK")
+label, _ = optional_import("skimage.measure", name="label")
+measure, _ = optional_import("skimage", name="measure")
+mesh, _ = optional_import("stl", name="mesh")
+resize, _ = optional_import("skimage.transform", name="resize")
+trimesh, _ = optional_import("trimesh")
+
+import monai.deploy.core as md
+from monai.deploy.core import DataPath, ExecutionContext, Image, InputContext, IOType, Operator, OutputContext
+
+__all__ = ["STLConversionOperator", "STLConverter"]
+
+
+@md.input("image", Image, IOType.IN_MEMORY)
+@md.output("stl_output", DataPath, IOType.DISK)
+@md.env(pip_packages=["numpy>=1.21", "numpy-stl>=2.12.0", "scikit-image>=0.17.2", "trimesh>=3.8.11"])
+class STLConversionOperator(Operator):
+    """Converts volumetric image to surface mesh in STL format, file output only."""
+
+    def __init__(
+        self, output_file=None, class_id=None, is_smooth=True, keep_largest_connected_component=True, *args, **kwargs
+    ):
+        """ "Creates an object to generate surface mesh in a STL file if the path is provided.
+
+        Args:
+            output_file (str, optional): output stl file relative path. Default to None for no file output.
+            class_id (array, optional): Class label ids. Defaults to None.
+            is_smooth (bool, optional): smoothing or not. Defaults to True.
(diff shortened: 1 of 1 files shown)


monai/deploy/operators/stl_conversion_operator.py
```diff
@@ -0,0 +1,443 @@
...
+@md.input("image", Image, IOType.IN_MEMORY)
+@md.output("stl_output", DataPath, IOType.DISK)
+@md.env(pip_packages=["numpy>=1.21", "numpy-stl>=2.12.0", "scikit-image>=0.17.2", "trimesh>=3.8.11"])
+class STLConversionOperator(Operator):
+    """Converts volumetric image to surface mesh in STL format, file output only."""
+
+    def __init__(
+        self, output_file=None, class_id=None, is_smooth=True, keep_largest_connected_component=True, *args, **kwargs
+    ):
+        """ "Creates an object to generate surface mesh in a STL file if the path is provided.
```
gigony, line 46:
```suggestion
        """Creates an object to generate a surface mesh in an STL file if the path is provided.
```

monai/deploy/operators/stl_conversion_operator.py
```diff
@@ -0,0 +1,443 @@
...
+
+        Args:
+            output_file (str, optional): output stl file relative path. Default to None for no file output.
```
gigony, line 49:
```suggestion
            output_file (str, optional): output STL file relative path. Default to None for no file output.
```

monai/deploy/operators/stl_conversion_operator.py
```diff
@@ -0,0 +1,443 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+import shutil
+import tempfile
+from pathlib import Path
+from typing import Dict, Union
```
gigony, line 17:
```suggestion
from typing import Dict, Optional
```

monai/deploy/operators/stl_conversion_operator.py
```diff
@@ -0,0 +1,443 @@
...
+            self._output_file = output_folder / self._output_file
+            self._output_file.parent.mkdir(exist_ok=True)
+            self._logger.info(f"Output will be saved in file {self._output_file}.")
+
+        stl_bytes = self._convert(input_image, self._output_file)
+
+        if not isinstance(op_output_config, DataPath):
+            op_output.set(stl_bytes)
+
+    def _convert(self, image: Image, output_file: Union[Path, None] = None):
```
gigony, line 95: it would be same with Optional.
```suggestion
    def _convert(self, image: Image, output_file: Optional[Path] = None):
```

monai/deploy/operators/stl_conversion_operator.py
```diff
@@ -0,0 +1,443 @@
...
+            is_smooth=self._is_smooth,
+            keep_largest_connected_component=self._keep_largest_connected_component,
+        )
+
+
+class STLConverter(object):
+    """Converts volumetric image to surface mesh in STL"""
+
+    def __init__(self, *args, **kwargs):
+        """ "Creates an instance to generate surface mesh in STL with an Image object."""
```
gigony, line 124:
```suggestion
        """Creates an instance to generate a surface mesh in STL with an Image object."""
```

monai/deploy/operators/stl_conversion_operator.py
```diff
@@ -0,0 +1,443 @@
...
+
+        super().__init__(*args, **kwargs)
+        self._logger = logging.getLogger("{}.{}".format(__name__, type(self).__name__))
+
+    def convert(
+        self,
+        image: Image,
+        output_file: Union[Path, None] = None,
```
gigony, line 132:
```suggestion
        output_file: Optional[Path] = None,
```

monai/deploy/operators/stl_conversion_operator.py
```diff
@@ -0,0 +1,443 @@
...
+                cube.vectors[i][j] = verts[f[j], :]
+
+        cube.save(os.path.splitext(filename)[0] + ".stl")
+
+    # Helper class for wrapping the App SDK Image object
+    #
+    class SpatialImage:
+        """Object encapsulating a spatial volume image instance of Image.
+
+        Channel is not support in this version.
```
gigony, line 266:
```suggestion
        Channel is not supported in this version.
```

monai/deploy/operators/stl_conversion_operator.py
```diff
@@ -0,0 +1,443 @@
...
+
+        def set_property(self, key: str, value):
+            """Sets an image property
+
+            Args:
+                key (str): key of the property
+                value: value of the property
+
+            Returns:
+
```
gigony, line 331: missing description for 'Returns'

dbericat:
+1 on using cuCIM where it makes sense!

MMelQin:
> I think the performance of this operation could be accelerated by using cuCIM (as label/resize methods in scikit-image are available in cuCIM) in the future.

Sounds good, though we'd allow both GPU and CPU for this particular operator, so cuCIM use will be optional/alternative.

PR #264, opened 2022-02-04
Corrected the conditional block to use the correct and only named input obj

MMelQin:

The conditional block was using a named input of DataPath type, but with a name that has been changed. In fact, during app execution, input of DataPath is not supported in the current SDK. Keeping this block in the code is forward thinking and in case the type of the input is modified to be DataPath. So, there had not been runtime issue in the released SDK versions.


gigony (approved):
Thanks Ming!
Looks good to me!

PR #265, opened 2022-02-11
Updated the png converter operator following changes to Image ndarray index order.

MMelQin:
The Image object that is created by the DICOM to Image converter now contains a numpy array with index order of DHW, the same way as ITK does. So, the depth (slicing along this axis for the original DICOM slice) is now the first dim, and the indexing needs to change accordingly.


MMelQin:
Thanks @mocsharp for the review. The change had also been tested with its built-in test method in dev env.

Issue #266, opened 2022-02-11
[BUG] DICOMDataLoaderOperator crashes when reading non-DICOM files


mocsharp:
**Describe the bug**
`DICOMDataLoaderOperator` crashes when trying to read a non-DICOM file.

**Steps/Code to reproduce bug**
```
! monai-deploy exec examples/apps/dicom_series_to_image_app/app.py -i [path-to-dicom-and-non-dicom-files] -o output

Traceback (most recent call last):
  File "/home/aaaaa/miniconda3/bin/monai-deploy", line 8, in <module>
   ...
  File "/home/aaaaa/miniconda3/lib/python3.8/site-packages/pydicom/filereader.py", line 623, in read_preamble
    raise InvalidDicomError("File is missing DICOM File Meta Information "
pydicom.errors.InvalidDicomError: File is missing DICOM File Meta Information header or the 'DICM' prefix is missing from the header. Use force=True to force reading.
```

**Expected behavior**
THe operator shall ignore and log non-DICOM files.

**Environment details (please complete the following information)**

- OS/Platform: linux, windows
- Python Version: 3.8.x
- Method of MONAI Deploy App SDK install: [pip, conda, Docker, or from source] source
- SDK Version: 0.2.1

PR #267, opened 2022-02-11
gh-266 handle and ignore non-dicom files

mocsharp:
Fix `DICOMDataLoaderOperator` to handle and ignore non-DICOM files.

### DIFF
--- a/monai/deploy/operators/dicom_data_loader_operator.py
+++ b/monai/deploy/operators/dicom_data_loader_operator.py
@@ -24,6 +24,7 @@ get_testdata_file, _ = optional_import("pydicom.data", name="dcmread")
 FileSet, _ = optional_import("pydicom.fileset", name="FileSet")
 generate_uid, _ = optional_import("pydicom.uid", name="generate_uid")
 valuerep, _ = optional_import("pydicom", name="valuerep")
+InvalidDicomError, _ = optional_import("pydicom.errors", name="InvalidDicomError")
 
 
 @md.input("dicom_files", DataPath, IOType.DISK)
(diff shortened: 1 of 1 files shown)


monai/deploy/operators/dicom_data_loader_operator.py
```diff
@@ -90,7 +91,10 @@ def _load_data(self, files: List[str]):
         sop_instances = []
 
         for file in files:
-            sop_instances.append(dcmread(file))
+            try:
```
MMelQin, line 94: It is good that we catch and suppress the exception on reading non-DICOM part10 files now, though this creates a chance of a silent failure when the given folder contains all non-DICOM files, for our use cases.
Consumers of this operator often check the loaded study list exists and not empty, and here is an [example](https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/64b40b71162eeb759b80f8a644c623133e85aec7/monai/deploy/operators/dicom_series_to_volume_operator.py#L46), but it would be better to raise an error in the compute method of this operator when nothing is loaded.

monai/deploy/operators/dicom_data_loader_operator.py
```diff
@@ -90,7 +91,10 @@ def _load_data(self, files: List[str]):
...
+                sop_instances.append(dcmread(file))
+            except InvalidDicomError:
+                print(f"Failed to read {file}; file may not be a valid DICOM file")
```
MMelQin, line 97: This can use logging.warning instead of print. Also, since the specific InvalidDICOMError is handled and not propagating up, it would be better to state the precise meaning of the error, "Failed to read a DICOM dataset from {file} because the dataset is not stored in accordance with the DICOM File Format."

monai/deploy/operators/dicom_data_loader_operator.py
mocsharp: Raise errors would just break unless there is a way to catch on the operator's level and be able to continue to a downstream operator.

MMelQin: That's exactly the point, error out at the point of not being able to load the necessary data, in this case, no DICOM study loaded, instead of letting the downstream consumers to raise error on missing required data,

PR #271, opened 2022-02-25
fix: Address typos in bilinear mode usage in Spacingd operator

swederik:
There is a typo in the use of the Spacing transform which causes some issues in the examples:

It's supposed to be `bilinear` instead of `blinear`

See:
https://github.com/Project-MONAI/MONAI/blob/05013c756db7b10ae9ac524cc2a7e68d75c6b693/monai/transforms/spatial/array.py#L354

Thanks!


dbericat:
Thanks @swederik ! Nice catch. LGTM! 

@MMelQin ?

MMelQin (approved):

MMelQin:
@swederik thank you for the catch and fix. Indeed, when "semi-automatically" creating the inference app, the inference config file was not copied over correctly. With the upcoming MONAI model package, transferring the inference config from training validation to deploy should be automated, one of its design goals is to avoid this type of error.

The following are the supported modes of SpacingD transform.
```
mode: {``"bilinear"``, ``"nearest"``}
                Interpolation mode to calculate output values. Defaults to ``"bilinear"``. 
```

PR #280, opened 2022-03-29
Fix pixel spacing in DICOMDataLoaderOperator for some X-rays

jlvahldiek:
We are working on a large dataset of single frame 2D x-ray images (DICOMs). Some of these images lack the DICOM tag "Pixel Spacing" (0028, 0030) resulting in a KeyError in DICOMDataLoaderOperator.

For some x-ray images (of our dataset but also in general) pixel spacing can only be stored in the DICOM tag "Imager Pixel Spacing" (0018, 1164). This PR adds code to cover this use case.

### DIFF
--- a/monai/deploy/operators/dicom_data_loader_operator.py
+++ b/monai/deploy/operators/dicom_data_loader_operator.py
@@ -261,7 +261,13 @@ class DICOMDataLoaderOperator(Operator):
                 series.row_pixel_spacing = pixel_spacing_de.value[0]
                 series.col_pixel_spacing = pixel_spacing_de.value[1]
         except KeyError:
-            pass
+            try:
+                imager_pixel_spacing_de = sop_instance[0x0018, 0x1164]
+                if imager_pixel_spacing_de is not None:
+                    series.row_pixel_spacing = imager_pixel_spacing_de.value[0]
+                    series.col_pixel_spacing = imager_pixel_spacing_de.value[1]
+            except KeyError:
+                    pass
 
         try:
             image_orientation_paient_de = sop_instance[0x0020, 0x0037]
--- a/monai/deploy/operators/dicom_data_loader_operator.py
+++ b/monai/deploy/operators/dicom_data_loader_operator.py
@@ -249,14 +249,14 @@ class DICOMDataLoaderOperator(Operator):
             pass
 
         try:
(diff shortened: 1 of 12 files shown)


MMelQin:
@jlvahldiek Thanks for catching this and providing the parsing logic for CR image.

In this [discussion](https://groups.google.com/g/comp.protocols.dicom/c/IerDfapgFYU?pli=1), David Clunie provided a historical perspective of the use of (0028, 0030) and (0018, 1164), and he suggested "Nowadays, a workstation has to look first for (0018,1164), if it doesn't find it look for (0028,0030), and if it doesn't find that look for whatever private attributes it is aware of."

Though, I think, as you coded up, we'll look for (0028, 0030) because CR Image module states "Physical distance in the Patient between the center of each pixel, specified by a numeric pair - adjacent row spacing (delimiter) adjacent column spacing in mm. See Section 10.7.1.1 and Section 10.7.1.3. Required if the image has been calibrated. May be present otherwise." while noting "Imager Pixel Spacing (0018,1164) is a required Attribute in DX family IODs.".

monai/deploy/operators/dicom_data_loader_operator.py
```diff
@@ -249,7 +249,14 @@ def populate_series_attributes(self, series, sop_instance):
             pass
 
         try:
-            pixel_spacing_de = sop_instance[0x0028, 0x0030]
+            tag_pixel_spacing = 'PixelSpacing'    # tag (0x0028, 0x0030)
+            tag_imager_pixel_spacing = 'ImagerPixelSpacing'    # tag (0x0018, 0x1164)
+            if tag_pixel_spacing in sop_instance: pixel_spacing_de = sop_instance[tag_pixel_spacing]
```
MMelQin, line 254: Code itself looks good to me, though the format checking fails and recommends "auto style fixes: ./run check --autofix", please see the build/test step, https://github.com/Project-MONAI/monai-deploy-app-sdk/runs/5761205272?check_suite_focus=true

MMelQin (approved):

monai/deploy/operators/dicom_data_loader_operator.py
MMelQin: Thanks.

PR #281, opened 2022-03-31
[WIP] Fix MONOCHROME1 Photometric Interpretation

jlvahldiek:
DICOM images having Photometric Interpretation of MONOCHROME1 are not inverted by monai-deploy-app-sdk.

The class DICOMSOPInstance reads in pixel_array via pydicom.dcmread(). Pydicom itself does not care about Photometric Interpretation and just delivers the stored pixel array (https://github.com/pydicom/pydicom/issues/783). 

I consider this behavior to be a bug (errorneous inverted images probably will have a huge impact on model's inference performance) and suggest a fix in this PR. Please tell me if there is a better place to put the pixel array inversion of MONOCHROME1 DICOMs.

With this fix DICOMDataloaderOperator reads in MONOCHROME1 images correctly inverted.

### DIFF
--- a/monai/deploy/core/domain/dicom_sop_instance.py
+++ b/monai/deploy/core/domain/dicom_sop_instance.py
@@ -43,6 +43,15 @@ class DICOMSOPInstance(Domain):
         return self._sop.__getitem__(key)
 
     def get_pixel_array(self):
+        try:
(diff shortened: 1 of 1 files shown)


jlvahldiek:
Maybe this issue should first be discussed if relevant and where to place the fix. Closing.

Issue #283, opened 2022-04-06
MONAI Seg Inference Operator Shared Memory


slbryson:
https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/8ec336fffffbaf54cbe3464540f4e36763d92360/monai/deploy/operators/monai_seg_inference_operator.py#L211

Applications that use the Dataloader in MONAI Seg Inference Operator throw errors when trying to allocate shared memory in Argo pipelines.

Need to modify the num of workers argument to not use shared memory by default.

MMelQin:
I changed this from `bug` to the closest label that I can find, because using shared_mem in dataloader is NOT a bug.

It had become known that when the application is packaged in a MAP (docker image) and some host platforms do not have the capability to set shared mem size on launching the MAP docker, e.g. Clara Deploy 0.8.1 (it is further noted that the default docker shared_mem size is 64MB while this can be modified in Docker daemon config), the MAP docker execution would fail on lack of shared_mem.

In order to allow the MAP docker to run even when the host platform does not bump the shared_mem size, shared_mem use is to be disabled as a workaround. Setting the num_workers to 0 may work per this [discussion](https://discuss.pytorch.org/t/can-we-disable-share-memory-usage-in-pytorch/47342).

slbryson:
I agree. It is not a bug, but I couldn't find a better label. Enhancement
would have been a better choice. Thank you.

aihsani:
I move to file another enhancement for MONAI app sdk to parameterize this via an environment variable or programmatically so that we may choose whether to use the feature or not. There's an obvious drawback to not using a variable number of workers.

MMelQin:
Thanks. Given the current use case, i.e. inference workflow limited to one image fully loaded before progressing, there is not much difference, but for sure configurability is definitely good.
We have created an issue to support MONAI Bundle in the next release with the highest priority, and this operator will likely see major changes, but the class API will be backward compatible as much as feasible. Will supply an optional parameter for num_workers on the constructor, with the same semantics as the one on torch dataloader.

aihsani:
Great, thank you

PR #284, opened 2022-04-06
Issue (#283) Fix shm issue

slbryson:
Signed-off-by: Sidney Bryson <sbryson@nvidia.com>


MMelQin (approved):
Looks good to me for the intended purpose of disabling shared memory use in the inference operator. Thanks!

I also tested with the AI Spleen Sseg App, Packager, Runner (with and without setting shm-size=1g on running MAP docker).

PR #288, opened 2022-04-08
Update MONAI version after having validated compatibility.

MMelQin:
Updated the version of monai (>=0.8.1) dependency after having verified the affected operators and applications work the same as with the older version monai(==0.6). This is expected since monai has not introduced breaking changes between these versions.


gigony (approved):
Thanks Ming! Looks good to me!

PR #290, opened 2022-04-13
Add support for MONOCHROME1

MMelQin:

Apply the "INVERSE" for pixel array with Phohometic Interpretation of MONOCHROME1/INVERSE, so that the minimum value is black, similar to MONOCHROME2/IDENTITY, during series to volume image conversion.

@jlvahldiek this is similar to what is in your aborted PR, but I have not compared the performance yet. Please review this PR. 

Note there are more enhancements needed, to come, e.g. considering the LUT, and compressed pixel data.

### DIFF
--- a/monai/deploy/operators/dicom_series_to_volume_operator.py
+++ b/monai/deploy/operators/dicom_series_to_volume_operator.py
@@ -110,13 +110,14 @@ class DICOMSeriesToVolumeOperator(Operator):
         photometric_interpretation = (
             slices[0].get_native_sop_instance().get("PhotometricInterpretation", "").strip().upper()
         )
+        presentation_lut_shape = slices[0].get_native_sop_instance().get("PresentationLUTShape", "").strip().upper()
 
         if not photometric_interpretation:
             logging.warning("Cannot get value of attribute Photometric Interpretation.")
 
         if photometric_interpretation != "MONOCHROME2":
-            if photometric_interpretation == "MONOCHROME1":
-                logging.debug(f"Inverting intensity of MONOCHROME1 image.")
+            if photometric_interpretation == "MONOCHROME1" or presentation_lut_shape == "INVERSE":
+                logging.debug("Applying INVERSE transformation as required for MONOCHROME1 image.")
                 vol_data = vol_data * -1 + np.max(vol_data)
             else:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


jlvahldiek:
Thank you, @MMelQin!

monai/deploy/operators/dicom_series_to_volume_operator.py
```diff
@@ -90,6 +95,35 @@ def generate_voxel_data(self, series):
...
+        )
+        presentation_lut_shape = slices[0].get_native_sop_instance().get("PresentationLUTShape", "").strip().upper()
+
+        if not photometric_interpretation:
+            logging.warning("Cannot get value of attribute Photometric Interpretation.")
+
+        if photometric_interpretation != "MONOCHROME2":
+            if photometric_interpretation == "MONOCHROME1" or presentation_lut_shape == "INVERSE":
+                logging.debug("Applying INVERSE transformation as required for MONOCHROME1 image.")
+                vol_data = vol_data * -1 + np.max(vol_data)
```
jlvahldiek, line 121:
```suggestion
                vol_data = np.amax(vol_data) - vol_data
```
In a recent project I used this code for inversions. Probably without a difference in performance.

monai/deploy/operators/dicom_series_to_volume_operator.py
```diff
@@ -90,6 +95,35 @@ def generate_voxel_data(self, series):
         vol_data = np.stack([s.get_pixel_array() for s in slices], axis=0)
         vol_data = vol_data.astype(np.int16)
```
jlvahldiek, line 96: In a recent project I used the following code in a subclass of DICOMSeriesToVolumeOperator in  generate_voxel_data() in order to apply the stored windowing (with good results):

`from pydicom.pixel_data_handlers.util import apply_voi_lut`

```
slices = series.get_sop_instances()
# The sop_instance get_pixel_array() returns a 2D NumPy array with index order
# of `HW`. The pixel array of all instances will be stacked along the first axis,
# so the final 3D NumPy array will have index order of [DHW]. This is consistent
# with the NumPy array returned from the ITK GetArrayViewFromImage on the image
# loaded from the same DICOM series.
mod_slices = []
for s in slices:
    mod_s = apply_voi_lut(s.get_native_sop_instance().pixel_array, s.get_native_sop_instance())
    mod_slices.append(mod_s)
vol_data = np.stack(mod_slices, axis=0)

vol_data = vol_data.astype(np.int16)
```

jlvahldiek: Should I contribute this in a separate PR?

monai/deploy/operators/dicom_series_to_volume_operator.py
MMelQin: Thank you. Updated per comment.

monai/deploy/operators/dicom_series_to_volume_operator.py
MMelQin: I think it is worth its own issue and corresponding PR, thanks.

gigony (approved):
Looks good to me!
Thank you!

PR #291, opened 2022-04-19
Add support for VOI LUT in DICOMSeriesToVolumeOperator

jlvahldiek:
- introduces the class attribute `_apply_voi_lut`
  - defaults to `False` preserving backwards compatibility
- uses pydicom's pixel data handler utilities to apply rescale operations and VOI LUT
  - `apply_modality_lut()` for rescale operation (changes former implementation)
  - `apply_voi_lut()` for windowing operation (optional, depends on class attribute `_apply_voi_lut`)

Successfully tested with the changes of #290

Adapted from: https://github.com/pydicom/pydicom/issues/1125

### DIFF
--- a/monai/deploy/operators/dicom_series_to_volume_operator.py
+++ b/monai/deploy/operators/dicom_series_to_volume_operator.py
@@ -84,7 +84,7 @@ class DICOMSeriesToVolumeOperator(Operator):
         return study_selected_series_list[0].selected_series[0].image
 
     def generate_voxel_data(self, series):
-        """Applies rescale slope and rescale intercept to the pixels.
+        """Applies rescale operation and also VOI LUT to the pixels, if required.
 
         Args:
             series: DICOM Series for which the pixel data needs to be extracted.
--- a/monai/deploy/operators/dicom_series_to_volume_operator.py
+++ b/monai/deploy/operators/dicom_series_to_volume_operator.py
@@ -32,6 +32,15 @@ class DICOMSeriesToVolumeOperator(Operator):
     Channel is limited to 1 as of now, and `C` is absent in the NumPy array.
     """
 
+    def __init__(self, apply_voi_lut: bool = False, *args, **kwargs):
+         """Creates an instance of this class
+         Args:
+             apply_voi_lut (bool): If true, tries to apply VOI LUT to voxel data.
+                                   Defaults to False.
+         """
+         super().__init__(*args, **kwargs)
+         self._apply_voi_lut = apply_voi_lut
+
     def compute(self, op_input: InputContext, op_output: OutputContext, context: ExecutionContext):
         """Performs computation for this operator and handles I/O."""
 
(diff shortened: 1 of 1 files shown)


monai/deploy/operators/dicom_series_to_volume_operator.py
```diff
@@ -30,6 +33,15 @@ class DICOMSeriesToVolumeOperator(Operator):
     Channel is limited to 1 as of now, and `C` is absent in the NumPy array.
     """
 
+    def __init__(self, apply_voi_lut: bool = False, *args, **kwargs):
```
MMelQin, line 36: Is VOI LUT always applied if present? Would a default of True for apply_voi_lut be better?

monai/deploy/operators/dicom_series_to_volume_operator.py
```diff
@@ -87,24 +99,16 @@ def generate_voxel_data(self, series):
...
+        for s in slices:
+            native_sop_instance = s.get_native_sop_instance()
+            # Apply rescale operation (if required, otherwise returns pixel_array unchanged)
+            rescaled_pixel_array = util.apply_modality_lut(native_sop_instance.pixel_array, native_sop_instance)
+            if self._apply_voi_lut:
+                # Apply windowing operation (if required, otherwise returns pixel_array unchanged)
+                rescaled_pixel_array = util.apply_voi_lut(rescaled_pixel_array, native_sop_instance)
+            pixel_arrays.append(rescaled_pixel_array)
+        vol_data = np.stack(pixel_arrays, axis=0)
         return np.array(vol_data, dtype=np.int16)
```
MMelQin, line 112: The line `return np.array(vol_data, dtype=np.int16)` now can potentially converts the pixel dtype (issue was there even before with the rescale being a large value), as both app_moality_lut and apply_voi_lut can return np.float64.

How best do we handle this, and how like the return value type would be np.float64? Should we at least test if np.float64, if true then return np with this type, else, return with conversion to np.int16? If the original pixel has type uint16, then it is still a problem. So would it either be np.float64, or int16 or uint16?  (original pixel data could be int16 while the other alternative returns from the two util functions are either uint 8 or uint16)

MMelQin:
@jlvahldiek Thanks for the PR. Was getting the App SDK release last Fri and have just added a couple comments. Please see if you can rebase the forked repo branch with the App SDK main.

Also, wonder if you have or have reference to a public dataset with DICOM instances with the attributes in question. I may also scan though the [TCIA ](https://www.cancerimagingarchive.net/) for test data.

Thanks.

jlvahldiek:
Thanks for your review, @MMelQin. Maybe this PR was a bit premature and i should have rather raised an issue with discussion first. I wonder if the DICOMSeriesToVolumeOperator should not load the pixel data analog to the MONAI core, i.e. with the ITKReader or monai.transforms.LoadImage().

Most models that will be used with monai-deploy-app-sdk have most likely been trained with the MONAI core. Therefore, in my opinion, the data loading in monai-deploy-app-sdk should be the same.

This would also mean that one would be able to abandon pydicom and thus not have to worry about setting modality lut, VOI lut or even Monochrome1/2.

I would like to close this PR and first start a discussion. What do you think?

monai/deploy/operators/dicom_series_to_volume_operator.py
jlvahldiek: In my opinion, only the modality LUT should always be used. This is in accordance with e.g. ITKReader.

Issue #87, opened 2021-08-30
Support multiple inputs/outputs from CLI


gigony:
Currently, we support one single file/folder path for the input and one folder path for the output.

Users may want to specify multiple inputs/outputs with labels.
We need to provide a way to specify and use multiple inputs/outputs.

vikashg:
@gigony Can you cite an example for this use case ?

PR #296, opened 2022-05-06
Fix the description of SingleProcessExecutor.run() method

gigony:
Existing description for SingleProcessExecutor.run() method is wrong.
This patch describes the correct behavior of the method.

### DIFF
--- a/monai/deploy/core/executors/single_process_executor.py
+++ b/monai/deploy/core/executors/single_process_executor.py
@@ -36,11 +36,11 @@ class SingleProcessExecutor(Executor):
     def run(self):
         """Run the app.
 
-        This method retrieves the root nodes of the graph traveres through the
-        graph in a depth first approach.
-        Retrieves output from an upstream operator at a particular output port.
-        Sets the right input to a downstrem operator at the right input port.
-        Executes the operators.
+        This method executes operators of the graph in topological order:
+
+        - If a node (operator) is a root node, its input is treated as a data path (DataPath, IOType.DISK).
+        - If a node (operator) is a leaf node, its output is treated as a data path (DataPath, IOType.DISK).
+        - After the execution of an operator, the output of the operator is used as the input of the next operator.
         """
         app_context = self.app.context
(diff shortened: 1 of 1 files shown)


monai/deploy/core/executors/single_process_executor.py
```diff
@@ -36,11 +36,11 @@ class SingleProcessExecutor(Executor):
...
         """Run the app.
 
-        This method retrieves the root nodes of the graph traveres through the
-        graph in a depth first approach.
-        Retrieves output from an upstream operator at a particular output port.
-        Sets the right input to a downstrem operator at the right input port.
-        Executes the operators.
+        This method executes operators of the graph in topological order:
+
+        - If a node (operator) is a root node, its input is treated as a data path (DataPath, IOType.DISK).
```
MMelQin, line 41: Multiple inputs supported?

monai/deploy/core/executors/single_process_executor.py
```diff
@@ -36,11 +36,11 @@ class SingleProcessExecutor(Executor):
...
+        - If a node (operator) is a leaf node, its output is treated as a data path (DataPath, IOType.DISK).
```
MMelQin, line 42: Multiple outputs (of the same type) supported?

MMelQin (approved):
Looks good to me, with just a couple clarification required comments.

monai/deploy/core/executors/single_process_executor.py
gigony: It is not supported yet and we have an issue regarding it https://github.com/Project-MONAI/monai-deploy-app-sdk/issues/87

All root nodes would be able to access only one input.

monai/deploy/core/executors/single_process_executor.py
gigony: It is not supported yet and we have an issue regarding it https://github.com/Project-MONAI/monai-deploy-app-sdk/issues/87

All leaf nodes would be able to access only one output.

Issue #195, opened 2021-10-27
[FEA] Include capabilities of highdicom to handle DICOM SEG and DICOM SR conversions 


GreySeaWolf:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

MMelQin:
A simple DICOM SR writer operator has been added as in the V0.2 release, in [PR](https://github.com/Project-MONAI/monai-deploy-app-sdk/pull/201), to support wrapping/reporting textual classification results.
This operator supports customizing the model info (used in the DICOM contributing equipment module), the equipment module, as well as DICOM attributes in a dict[str, str] for setting attribute with string based VR or which be converted from str.

Also created was a DICOM Encapsulated PDF writer for the same purpose, though I did not merge it.
DICOM RT Struct and STL are also planned to be ported over to MONAI (developed in Clara and have been in used at NIH).

The current DICOM Seg writer in the App SDK is expected to be refactored; much of the code was contributed from Brad Genereaux a while back, and I'm still yet to start the the refactoring that had been planned.

CPBridge:
Current status:

- We have SEG
- We have basic text SR, not using highdicom
- We still need Comprehensive(3D)SR using the TID1500 template

PR #298, opened 2022-05-15
Introduce highdicom library for DICOM Segmentations

CPBridge:
Addresses #195 (partially)

Introduces the highdicom library to handle the creation of DICOM segmentations. This should lead to more maintainable code by using the well-tested highdicom package for the complex task of segmentation creation, and should make adding new features to the segmentation writer operator considerably easier by leveraging the prior work in that library. Note that the complexity of the ad-hoc code in the dicom segmentation operator file as reduced signficantly.

I am considering this a largely complete "minimum viable" implementation and would appreciate feedback at this stage about whether we should go forward with this and what may need to change. Note that this is a backwards incompatible change to the signature of the `DICOMSegmentationWriterOperator`'s `__init__` method. It is now necessary to pass a list of `highdicom.seg.SegmentDescription` objects to the constructor to describe each segment in the created segmentation. I'd argue that the existing implementation, which hard codes in some default values, is likely to create dicom segs that are semantically incorrect in many cases (e.g. hard coding the segmented category as organ when it may be an abnormality such as a tumor or nodule). I am assuming that this backwards incompatibility is acceptable at this stage since the docstring states that the interface may change. Would appreciate thoughts on this.

I have attempted to update the two examples that use the seg writer: spleen and liver tumor segmetentations. I have managed to test the spleen example and visualize the results, and it's working nicely. I have not managed to test the liver tumor example. I may be being stupid but I can't find where to download the relevant files to actually run it.

Another thing I'm a bit confused about is where and how to add highdicom as an (optional) requirement. There are lots of requirements files and I wasn't totally sure how it all fits together. For now I just added highdicom to requirements_dev.txt. Please let me know whether this is correct.

My apologies that this took much longer than I anticipated to make happen (I largely blame [this monster](https://github.com/herrmannlab/highdicom/pull/139)). Now I've figured out how this stuff all fits together, adding further operators (such as SR, GSPS) using the capabilities in highdicom should become much easier if we choose to merge this.

Tagging highdicom author and co-maintainer @hackermd for visibility.


CPBridge:
Update: apparently I cannot type hint a variable with a highdicom type if highdicom is an optional import, which is why builds are currently failing. We will need to think about either making highdicom a required dependency, or come up with another way to pass all the relevant information to the DICOMSegmentationWriterOperator's constructor

MMelQin:
> Update: apparently I cannot type hint a variable with a highdicom type if highdicom is an optional import, which is why builds are currently failing. We will need to think about either making highdicom a required dependency, or come up with another way to pass all the relevant information to the DICOMSegmentationWriterOperator's constructor

Thanks for the PR.

There had been a architecture decision at the onset of the App SDK that there should be minimal number of dependencies in the App SDK proper, much like the MONAI core, and the same technique of using optional import is carried over to App SDK. In fact, even monai package itself is not a dependency of App SDK itself, although a few built-in operators/classes have to optionally import monai.

There had been an issue with the optional import in the built-in operator, MonaiSegInferenceOperator, which caused runtime error when the operator is not used in the application and rightfully the application does not include the optional depends in its requirements, but the issue was addressed in v0.2.

CPBridge:
Thanks a lot @gigony that seems to have fixed it nicely

MMelQin:
@CPBridge Can we add this new implementation of DICOM Seg writer as a new class/operator for v0.4? This way we allow a gradual transition and options for the users, and the example applications will not need to change for v0.4?

CPBridge:
Closing in favour of #327

Issue #302, opened 2022-06-08
mednist_app:latest -- No module named 'nibabel'


dbericat:
### Discussed in https://github.com/Project-MONAI/monai-deploy-app-sdk/discussions/301

<div type='discussions-op-text'>

<sup>Originally posted by **Leengit** June  8, 2022</sup>
I am trying to run the Jupyter lab notebook [02_mednist_app.ipynb](https://docs.monai.io/projects/monai-deploy-app-sdk/en/latest/_static/notebooks/tutorials/02_mednist_app.ipynb) on my local machine but the invocation of
```bash
!monai-deploy run mednist_app:latest input output
```
is giving me `ModuleNotFoundError: No module named 'nibabel'`.  Thoughts?

To reproduce:
```bash
sudo apt install cuda=11.3.1-1 libcudnn8-dev=8.2.1.32-1+cuda11.3 libcudnn8=8.2.1.32-1+cuda11.3
python3 -m venv ~/venv/mednist
source ~/venv/mednist/bin/activate
pip install -U pip setuptools wheel ipykernel
python -m ipykernel install --user --name=mednist
pip install -U torch torchvision torchaudio ipywidgets 'monai[pillow, tqdm, ignite, gdown, nibabel]' \
    monailabel monai-deploy-app-sdk pandas transformers mlflow \
    --extra-index-url https://download.pytorch.org/whl/cu113
# Then run the whole notebook using the `mednist` kernel
```

The full output from `!monai-deploy run mednist_app:latest input output` is
```text
Checking dependencies...
--> Verifying if "docker" is installed...

--> Verifying if "mednist_app:latest" is available...

Checking for MAP "mednist_app:latest" locally
"mednist_app:latest" found.

Reading MONAI App Package manifest...
--> Verifying if "nvidia-docker" is installed...

Traceback (most recent call last):
  File "/opt/monai/app/mednist_classifier_monaideploy.py", line 13, in <module>
    import monai.deploy.core as md  # 'md' stands for MONAI Deploy (or can use 'core' instead)
  File "/root/.local/lib/python3.8/site-packages/monai/__init__.py", line 43, in <module>
    load_submodules(sys.modules[__name__], False, exclude_pattern=excludes)
  File "/root/.local/lib/python3.8/site-packages/monai/utils/module.py", line 187, in load_submodules
    mod = import_module(name)
  File "/opt/conda/lib/python3.8/importlib/__init__.py", line 127, in import_module
[...]

PR #308, opened 2022-07-06
Mqin/fix mednist app error missing nibabel pkg

MMelQin:
Per reported issue https://github.com/Project-MONAI/monai-deploy-app-sdk/issues/302, the mednist MAP fails because of missing nibabel package, all because of the non-"optional import" for it in the stl_conversion_operator.py.

The mednist application does not use STLConversionOperator, whose dependencies are therefore not picked up by the App's info gathering through the env decorator. Not setting optional import then causes missing package issues when the not satisfied package is not set as optional import.

### DIFF
--- a/monai/deploy/operators/stl_conversion_operator.py
+++ b/monai/deploy/operators/stl_conversion_operator.py
@@ -36,6 +36,7 @@ __all__ = ["STLConversionOperator", "STLConverter"]
 
 @md.input("image", Image, IOType.IN_MEMORY)
 @md.output("stl_output", DataPath, IOType.DISK)
+# nibabel is required by the dependent class STLConverter.
 @md.env(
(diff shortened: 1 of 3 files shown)


gigony (approved):
Thanks @MMelQin !

Looks great to me!

MMelQin:
Thanks Vikash for reviewing it. 

It looks the previous version of the code was still present when the app was run, for some reason, specifically this entry reveals it was the buggy version of the code, 
```
File "/home/gupta/anaconda3/envs/monaideploy2/lib/python3.8/site-packages/monai/deploy/operators/stl_conversion_operator.py", line 19, in <module>
    import nibabel as nib
```

Can you please check in your env the updated code is present? Specifically `import nibabel as nib` should have been gone, and the `nib, _ = optional_import("nibabel")` appears on [line 23](https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/main/monai/deploy/operators/stl_conversion_operator.py#L23).

PR #313, opened 2022-07-12
Add the newly introduced example to readme.md

MMelQin:
This is a a bit of a chicken-and-egg problem, since the readme.md refers
to the published latest version of the doc which can only happen when
the release build is done, and this particular doc is part of the build,
included in the package to be pushed to pypi.org.

Next time, need to consider add the entry, with a placeholder/non-existent
URL prior to the release build, so that the content will show up in the new
package, and URL leads to valid content post build.


gigony (approved):
Looks good to me!

PR #316, opened 2022-07-26
DOC: Use upstream/downstream instead of source/destination

Leengit:
For monai.deploy.core.Operator, instead of sometimes using `upstream` and `downstream` and sometimes using `source` and `destination`, consistently use the former.


MMelQin:
@Leengit Thank you very much of the pull request.

The use of `source` and `destination` is consistent in the App SDK core, while the example apps, which are not included in the monia-deploy-app-sdk PyPi package, use `upstream` and `downstream`.

I believe `source` and `destination` are more precise, especially considering when using the `add_flow` function to connect two operators, a `source` is connected to a `destination`.  So, I'd say that `upstream` and `downstream` in the example apps be changed to `source` and `destination` respectively.

Leengit:
@MMelQin I have done as you suggested with this latest commit.

MMelQin (approved):
Thank you for the changes to have consistent use of naming.

MMelQin:
The build failure is caused by the style checking tool flake8 itself, which was working fine. 
  `ModuleNotFoundError: No module named 'flake8.plugins.manager'`
  
 The issue with the tool itself will be looked at, and the PR itself can be merged.

PR #327, opened 2022-08-15
Implement highdicom seg operator

CPBridge:
I am recreating this pull request since I completely borked #298 due to the (extremely frustrating) DCO sign off requirement and subsequent attempts to fix it that made it worse :persevere: 

This PR introduces the highdicom library to handle the creation of DICOM segmentations. This should lead to more maintainable code by using the well-tested highdicom package for the complex task of segmentation creation, and should make adding new features to the segmentation writer operator considerably easier by leveraging the prior work in that library. Note that the complexity of the ad-hoc code in the dicom segmentation operator file is reduced significantly.

Unlike the original version of #298 I have now removed the highdicom class `highdicom.seg.SegmentDescription` from the public API of the monai deploy class. However I have left the `pydicom.sr.coding.Code` class in the public API as we discussed in the last meeting.

I have not yet changed the examples to use the new version of the operator, as this requires merging with some recent changes due to the bundle that I still need to get my head around. This will need to be addressed before merging so that the examples run. But I would like to get agreement on the API of the class before fixing the examples.

Update:
All example apps that make use of the Segmentation Writer have been updated in this PR, and all used the appropriate codes. Also the monai core is pinned at v0.9.0 because the latest version 0.9.1 is not compatible with the apps (not related to the new Seg Writer itself)

### DIFF
--- a/monai/deploy/operators/dicom_seg_writer_operator.py
+++ b/monai/deploy/operators/dicom_seg_writer_operator.py
@@ -9,16 +9,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
-import datetime
-import json
-import logging
 import os
 from pathlib import Path
 from random import randint
-from typing import List, Optional, Union
+from typing import List, Optional, Sequence, Union
 
 import numpy as np
+from typeguard import typechecked
 
 from monai.deploy.utils.importutil import optional_import
 from monai.deploy.utils.version import get_sdk_semver
@@ -28,8 +25,10 @@ generate_uid, _ = optional_import("pydicom.uid", name="generate_uid")
 ImplicitVRLittleEndian, _ = optional_import("pydicom.uid", name="ImplicitVRLittleEndian")
 Dataset, _ = optional_import("pydicom.dataset", name="Dataset")
 FileDataset, _ = optional_import("pydicom.dataset", name="FileDataset")
-Sequence, _ = optional_import("pydicom.sequence", name="Sequence")
+hd, _ = optional_import("highdicom")
 sitk, _ = optional_import("SimpleITK")
+codes, _ = optional_import("pydicom.sr.codedict", name="codes")
+Code, _ = optional_import("pydicom.sr.coding", name="Code")
 
 import monai.deploy.core as md
 from monai.deploy.core import DataPath, ExecutionContext, Image, InputContext, IOType, Operator, OutputContext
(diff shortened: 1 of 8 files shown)


MMelQin:
@CPBridge Thank you very much for the updated PR.

I was very eager to try it out with the Spleen example app. Happy to say that the new writer works well, after I fixed a couple issues in other relevant sources,

1. Not an issue itself but expected change: the App needs to be updated to import pydicom for the `codes`, as well as importing and instantiating the `SegmentDescription`, as you did with the test function. The original impl has a note stating the Seg Writer `seg_labels` argument was subject to change, so it is fine even with this breaking change.
2. MONAI 0.9.1 `MetaTensor` broke the `InvertD` function used in the post-transform, so the re-sampled predicted image does not get inverted back to the original pixel spacing then slices. This caused the exception of mismatched slices of pred and original image, as expected. This not an issue when MONAI 0.9.0 is used. Also, there is already an issue to address the compatibility  with 0.9.1.
3. HighDicom is not handling the exception on DA conversion from PyDicom, e.g. the original DICOM series, generated with 3D Slicer, has the `StudyDate` of `2019-09-16`, not conformant to DICOM DA VR type, so PyDicom throws and so does HighDicom. The existing Seg Writer is more tolerant by simple reusing the original DA content, as such the problematic series has been in use in the public test dataset. I ran a different series with conformant headers successfully completed the test. I will locate/create and publish a conformant series for the test dataset.

How many segments can the new writer support? The existing one can support upto 8, as it caps the BitsAllocated at 8, as such the UNETR example app could not use the exiting Seg Writer for it can segment upto 13 types of organs.

CPBridge:
Thanks for trying it out @MMelQin !

Would it make sense for you to push your fixes for the spleen segmentation app directly to this PR?

> HighDicom is not handling the exception on DA conversion from PyDicom, e.g. the original DICOM series, generated with 3D Slicer, has the StudyDate of 2019-09-16, not conformant to DICOM DA VR type, so PyDicom throws and so does HighDicom. The existing Seg Writer is more tolerant by simple reusing the original DA content, as such the problematic series has been in use in the public test dataset. I ran a different series with conformant headers successfully completed the test. I will locate/create and publish a conformant series for the test dataset.

From the start, we have keen that highdicom should not contribute to the huge amount of non-compliant DICOM around, so we have implemented quite strict checks on input values to ensure that any file we produce is valid (and deliberately propagate some errors up from pydicom too, as you found). This does occasionally cause issues with existing source data that is invalid, such as this date formatting issue. Similarly the test data used for the unit test has an invalid name (`"John Doe"` rather than `"Doe^John"`) which raises a warning in that case. In my opinion the best way to fix this is to fix the test data as you suggest. If we try and work around every possible way the input data we could be invalid, it would be an impossible task.

> How many segments can the new writer support? The existing one can support upto 8, as it caps the BitsAllocated at 8, as such the UNETR example app could not use the exiting Seg Writer for it can segment upto 13 types of organs.

If this is true then I think that the current implementation might be quite wrong. I believe per the standard, BtisAllocated is capped at 8 for the Segmentation IOD but the BitsAllocated is irrelevant to the number of classes since each segment is recorded as its own frame within the segmentation. (The only use for all 8 bits is when using "fractional" segmentations, where a pixel value represents either the probability that a pixel belongs to a segment, or a the occupancy of that pixel that belongs to the segment. Highdicom supports both but I have not yet implemented fractional segmentations in the seg writer but will try to add this once the overall approach is agreed upon). Therefore there is in principle no limit on the number of segments, and highdicom has no limit in theory. In practice of course a large number of segments can be quite space intensive. Any way long story short this should be no problem at all to use for the 13 classes of the UNETR example, I have personally used the highdicom implementation for more than that in the past.

So are you happy with the overall approach and API? If so I will make a couple of further generalizations and fix up the spleen segmentation app before merging?

MMelQin:
@CPBridge Thanks for the clarification.
I have seen DICOM series manually generated, from Medical Decathlon dataset, by users with 3D Slicer, which is very tolerant on non-conformant VR, often times with flipped images too, hence took the approach of directly copying the necessary attributes. I am all for proper validation even when reusing attributes, since with the use of TCIA dataset there would be less of a chance getting non-conformant series.

The existing Seg Writer supports only Binary Seg type, and has the segment sequence properly assigned. "As a consequence of the enumerated Bits Allocated and Bits Stored Attribute values, single bit pixels shall be packed 8 to a byte as defined by the encoding rules in [PS3.5](http://dicom.nema.org/medical/dicom/current/output/html/part05.html#PS3.5).", so the pixel value binary is quite compact. In any case, I will for sure try out the UNETR app with >8 segments.

CPBridge:
> As a consequence of the enumerated Bits Allocated and Bits Stored Attribute values, single bit pixels shall be packed 8 to a byte as defined by the encoding rules in [PS3.5](http://dicom.nema.org/medical/dicom/current/output/html/part05.html#PS3.5).

I think I see the possible root of the confusion relating to **which** 8 pixel values are "packed" into a byte. For segmentations with SegmentationType of "BINARY", the BitAllocated is always 1 regardless of the number of segments, meaning that there are 8 pixel values stored within a single byte. However, these 8 pixels are not drawn from 8 different segments but rather from **8 spatially consecutive pixels within a given segment** (you can see this since pixels are encoded as Pixel Sample Values from [left to right and top to bottom](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/chapter_D.html) and for binary segmentations there is only one value per Pixel Sample Value since [SamplesPerPixel is 1](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.2.html#sect_C.8.20.2.1)). Each segment is encoded by a separate frame of the segmentation, and there can be an arbitrary number of these. As a consequence there are still 8 pixels per byte but no limit on the number of segments.

If the old implementation was packing 8 pixels from different segments together into a byte (I haven't checked the implementation to see whether or not it was) then I think it is likely wrong.

monai/deploy/operators/dicom_seg_writer_operator.py
```diff
@@ -157,12 +262,29 @@ def create_dicom_seg(self, image: np.ndarray, dicom_series: DICOMSeries, file_pa
...
+        seg = hd.seg.Segmentation(
+            source_images=dicom_dataset_list,
+            pixel_array=image,
+            segmentation_type=hd.seg.SegmentationTypeValues.BINARY,
+            segment_descriptions=self._seg_descs,
+            series_instance_uid=hd.UID(),
+            series_number=random_with_n_digits(4),
+            sop_instance_uid=hd.UID(),
+            instance_number=1,
+            manufacturer="The MONAI Consortium",
```
MMelQin, line 280: Nit: For now this set of attributes can be left hard coded. They are mostly used for the contributing equipment module,  and have been encapsulated in the App SDK class, `EquipmentInfo`, which is still mixed in the module of the SR writer. The plan was to have the Model Info and Equipment Info all encapsulated and passed down from the App to all writer types, so that the generated DICOM objects have correct attributes conforming to the IHE AI Result Profile. profile, etc.

MMelQin:
> If the old implementation was packing 8 pixels from different segments together into a byte (I haven't checked the implementation to see whether or not it was) then I think it is likely wrong.

Yes, for sure needed a look when we have time. That piece of code was contributed by a guru on the DICOM Standards Committees, quite a number of years back. I looked into the issues with bits packing for more than 8 segments, but put it in the back burner due to other high priority items, and also the multi-seg DICOM Seg could be loaded and viewed in OHIF Viewer. The other issue was that existing impl hardcoded the Recommended Display CIELab Value Attribute to the same value for all segments; on the plan to fix too, but now the new impl does not set this Type 3 attribute, so one less concern.

@CPBridge I have updated the app, and pinned the monai to v0.9.0 in the requirements txt as well as in the App SDK operators, so that both local run as well as Docker run succeed. Switching back to >=0.9.0 will be done when the combability issue is addressed. Please review and approve these changes.

MMelQin (approved):
Look good to me. The updated example App runs successfully in local dev mode as well as in Docker, and the output DICOM Seg instance was inspected for metadata correctness as well as image display with MicroDicom.

MMelQin:
@CPBridge I have found all the codes for the 13 organs in the UNETR app, by searching the pydicom dictionary modules and SNOMED CT portals. So I am happy with the apps now. The other nits on filling in more useful but non mandatory attributes, e.g. series description, date and time, contributing equipment module, etc. can be added later in separate lower priority issues.
I will go ahead and finally merge this PR. It is a good one, and definitely we can consider to expand and include the other writers. Thanks.

CPBridge:
Hey @MMelQin thanks so much for the review and for bringing the examples in line with the changes! Glad it's working well. Now we have figured out how this can work, I can work on other operators leveraging the things we have developed in the library (various types of SR, GSPS, Parametric Map)

MMelQin:
> Hey @MMelQin thanks so much for the review and for bringing the examples in line with the changes! Glad it's working well. Now we have figured out how this can work, I can work on other operators leveraging the things we have developed in the library (various types of SR, GSPS, Parametric Map)

@CPBridge Thank you too for having a good design and bring this key package into App SDK, without much learning curve for our target user persona. This is definitely the beginning a good integration, and happy to see the good output will circle back and benefit the production deployment is the clinics.

requirements-examples.txt
```diff
@@ -8,4 +9,4 @@ nibabel >= 3.2.1
 numpy-stl >= 2.12.0
 trimesh >= 3.8.11
 torch >= 1.10.0
-monai >= 0.9.0
\ No newline at end of file
+monai == 0.9.0
```
wyli, line 12: are there any specific issues working with monai v0.9.1? we can improve the core codebase...

MMelQin: The v0.9.1 broke the use of invertd with the way it is used in the apps, I have created an issue to update the app and will consult the migration guide. Thanks.

PR #328, opened 2022-08-16
Nuance PIN MONAI Integration Example App

aihsani:
Creating an example integration app to allow MONAI users to run existing MONAI apps in Nuance's PIN network.

This integration is currently only for users/developers with access to the Nuance PIN SDK.


dbericat:
Love this PR!

I'd like to place it somewhere else though, rather than /examples. 

I think it would be better to have an /integrations folder. Then a subfolder with /marketplaces/nuance_pin . Then another subfolder with other OSS communities like medperf and the linux foundation. Etc.

These integrations are key and hugely invaluable for the success of the SDK. I feel like putting them under /examples hide them and lower the value. 

Thoughts? @aihsani @MMelQin @rahul-imaging @gigony @brad @CPBridge

MMelQin:
> Love this PR!
> 
> I'd like to place it somewhere else though, rather than /examples.
> 
> I think it would be better to have an /integrations folder. Then a subfolder with /marketplaces/nuance_pin . Then another subfolder with other OSS communities like medperf and the linux foundation. Etc.
> 
> These integrations are key and hugely invaluable for the success of the SDK. I feel like putting them under /examples hide them and lower the value.
> 
> Thoughts? @aihsani @MMelQin @rahul-imaging @gigony @brad @CPBridge

I discussed with Alvin earlier on where to place this adaptation code, and a folder under `examples` seemed to be a good candidate for now. The main reason is that this piece is not in the App SDK PyPi package proper, at for now, so it live only live in the repo, as an "example"

We can consider and discuss along the lines of extensions, integrations, applications etc., and rename/reshuffle the `examples` and `notebook` folder if need be.

MMelQin:
I've also created a discussion topic on integration maturity level, #330. A good organization of folders will help a reader to navigate, but still have limitations without properly addressing use case in the App SDK proper.

examples/integrations/nuance_pin/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+# run AiSvcTest with spleen dicom payload
+python -m AiSvcTest -i ~/Downloads/dcm -o ~/Downloads/dcm/out -s http://localhost:5000 -V 2 -k
+```
+
+### Bring Your Own MONAI App
+
+This example integration may be modified to fit any existing MONAI app, however, they may be caveats.
+
+Nuance PIN requires all artifacts present in the output folder to be also added into the `resultsManifest.json` output file
```
CPBridge, line 156: Is it `resultsManifest.json` or `resultManifest.json` (s or no s), both are used here?

aihsani:
> > Love this PR!
> > I'd like to place it somewhere else though, rather than /examples.
> > I think it would be better to have an /integrations folder. Then a subfolder with /marketplaces/nuance_pin . Then another subfolder with other OSS communities like medperf and the linux foundation. Etc.
> > These integrations are key and hugely invaluable for the success of the SDK. I feel like putting them under /examples hide them and lower the value.
> > Thoughts? @aihsani @MMelQin @rahul-imaging @gigony @brad @CPBridge
> 
> I discussed with Alvin earlier on where to place this adaptation code, and a folder under `examples` seemed to be a good candidate for now. The main reason is that this piece is not in the App SDK PyPi package proper, at for now, so it only lives in the repo, as an "example"
> 
> We can consider and discuss along the lines of extensions, integrations, applications etc., and rename/reshuffle the `examples` and `notebook` folder if need be.

I've updated the location to `/integration` to increase profile

PR #337, opened 2022-08-31
Improve STL operator so the enclosing app obj can run repetitively

MMelQin:
In this PR, the STL Conversion has been improved
- execution of `compute` will not affect the the object state, and only input data will be transformed to be the output
- Fix the output to be in_memory STL binary in a try/catch block, so this output will only be available when the operator is a non-leaf node (current limitation on output mapping)
- STL file is only saved in the app's output folder, if and only if a file path is provided, which is stored as operator private attribute

Changes are also made to the unetr example to demonstrate reusing the same app obj for processing multiple/batch of inputs,
- `__main__` functions have been updated to process a set of subfolders under the app input folder, each containing a DICOM Study/series, to demonstrate batch processing. Example code can be customized to be run with a main loop which gets discrete input with the use of file mon, queue, or other triggers like interrupt handler.

Example output (note, the input DICOM study was duplicated into separate folder, hence the same series instance UID appears in all output folders),
```
output
├── 1-SelectedSeriesForUNETR_output
│   ├── dicom_seg-DICOMSEG.dcm
│   ├── publish
│   │   ├── 1.2.826.0.1.3680043.2.1125.1.nii.gz
│   │   ├── 1.2.826.0.1.3680043.2.1125.1_seg.nii.gz
│   │   ├── config.meta
│   │   └── config_render.json
│   └── stl
│       └── multi-organs.stl
├── 2-SelectedSeriesForUNETR_output
│   ├── dicom_seg-DICOMSEG.dcm
│   ├── publish
│   │   ├── 1.2.826.0.1.3680043.2.1125.1.nii.gz
│   │   ├── 1.2.826.0.1.3680043.2.1125.1_seg.nii.gz
│   │   ├── config.meta
│   │   └── config_render.json
│   └── stl
│       └── multi-organs.stl
└── 3-SelectedSeriesForUNETR_output
    ├── dicom_seg-DICOMSEG.dcm
    ├── publish
    │   ├── 1.2.826.0.1.3680043.2.1125.1.nii.gz
    │   ├── 1.2.826.0.1.3680043.2.1125.1_seg.nii.gz
    │   ├── config.meta
    │   └── config_render.json
    └── stl
        └── multi-organs.stl
```

### DIFF
--- a/examples/apps/ai_unetr_seg_app/__main__.py
+++ b/examples/apps/ai_unetr_seg_app/__main__.py
@@ -30,6 +30,8 @@ if __name__ == "__main__":
     for idx, dir in enumerate(input_dirs):
         try:
             output_path = app_output_path / f"{dir.name}_output"
+            # Note: the work_dir should be mapped to the host drive when used in
+            #       a container for better performance.
             work_dir = f".unetr_app_workdir{idx}"
             work_dirs.extend(work_dir)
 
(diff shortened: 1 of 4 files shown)


MMelQin:
> Thanks Ming! Looks good to me.

Thanks @gigony !

PR #345, opened 2022-09-13
Added support of matching instance level tags, and update liver seg

MMelQin:
Existing series selection implementation supports matching attributes at down to series level, however, sometimes key instance level attributes can help narrow down the selection, e.g. `Photometric Interpretation`.

Also updated the example app for make use of the updated selector, as well as pinning the monai dependency at 0.9.0 for now.
   
### DIFF
--- a/examples/apps/ai_livertumor_seg_app/app.py
+++ b/examples/apps/ai_livertumor_seg_app/app.py
@@ -21,7 +21,6 @@ from monai.deploy.operators.dicom_data_loader_operator import DICOMDataLoaderOpe
 from monai.deploy.operators.dicom_seg_writer_operator import DICOMSegmentationWriterOperator, SegmentDescription
 from monai.deploy.operators.dicom_series_selector_operator import DICOMSeriesSelectorOperator
 from monai.deploy.operators.dicom_series_to_volume_operator import DICOMSeriesToVolumeOperator
-
 from monai.deploy.operators.publisher_operator import PublisherOperator
 
 # This is a sample series selection rule in JSON, simply selecting CT series.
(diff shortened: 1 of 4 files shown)


monai/deploy/operators/dicom_series_selector_operator.py
```diff
@@ -219,7 +219,17 @@ def _select_series(self, attributes: dict, study: DICOMStudy, all_matched=False)
...
+                logging.info(f"    Series attribute {key} value: {attr_value}")
+
+                # If not found, try the best at the native instance level for string VR
+                # This is mainly for attributes like ImageType
+                if not attr_value:
+                    try:
+                        attr_value = [series.get_sop_instances()[0].get_native_sop_instance()[key].repval]
+                        series_attr.update({key: attr_value})
+                    except Exception:
+                        logging.info(f"        Attribute {key} not at instance leve either.")
```
gigony, line 231:
```suggestion
                        logging.info(f"        Attribute {key} not at instance level either.")
```

examples/apps/ai_livertumor_seg_app/app.py
```diff
@@ -21,8 +21,30 @@
 from monai.deploy.operators.dicom_seg_writer_operator import DICOMSegmentationWriterOperator, SegmentDescription
 from monai.deploy.operators.dicom_series_selector_operator import DICOMSeriesSelectorOperator
 from monai.deploy.operators.dicom_series_to_volume_operator import DICOMSeriesToVolumeOperator
+
```
gigony, line 24:
```suggestion
```

gigony (approved):
Looks good to me!

monai/deploy/operators/dicom_series_selector_operator.py
MMelQin: Thanks. Fixed

examples/apps/ai_livertumor_seg_app/app.py
MMelQin: Thanks. Fixed.

PR #346, opened 2022-09-15
Enhance some attributes in the text SR writer

MMelQin:
This pull request intends to merge some changes made during the IHE Connectathon IAIP testing.
- Update the series selection operator to support matching DICOM instance level metadata
- Updated the liver tumor segmentation example to include simple filtering rules
- Updated the text SR writer to add time zone offset from UTC, while keep content date time local
- Updated the text SR writer to use content sequence for sections of text
- Updated the text SR writer storage SOP class to that of basic text SR


vikashg:
Everything looks good to me.

MMelQin:
> Thanks for adding Media Storage SOP Class UID, that was really helpful @MMelQin

The Storage SOP Class UID attributed was changed from Comprehensive 3D to the more representative Basic Text SR in the PR.

Also, I need to reopen this to merge the PR.

PR #360, opened 2022-10-01
App SDK updated to be compatible with monai 1.0.0 and its MetaTensor

MMelQin:
The pull request is to make App SDK compatible with monai 1.0.0, specifically the use of MetaTensor
- Removed the use of `ToTensorD` in the inference pre-processing step
- Updated the InMemImageReader to make use the newly available `MetaKeys` class etc
- Tested the example apps successfully with pip installed monai 1.0.0
- Update the bundle inference operator to pass keyed input MetaTensors, e.g. `image`, to post_processing, since the metadata dict is now encapsulated in the MetaTensors
- Updated the Jupyter Notebooks. NOTE: testing Clara-Viz notebook needs to `pip install ipywidgets==7.7.2` before the clara-viz package is updated to be compatible with newer `ipywidgets` 
- Updated the PIN integration code where monai transforms are used
- Updated the Seg Writer to add Series Date, Time, and UTC time zone offset, as well as adding support for setting custom tags, e.g. SeriesDescription as it is missing from highdicom lib.


vikashg (commented):
@MMelQin Do we have the sample data for liver tumor somewhere that we can test.

vikashg (changes requested):
For the notebook tutorials I am getting the following error. 
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [6], line 1
----> 1 app = AISpleenSegApp()
      3 app.run(input="dcm", output="output", model="model.ts")

Cell In [5], line 4, in AISpleenSegApp.__init__(self, *args, **kwargs)
      3 def __init__(self, *args, **kwargs):
----> 4     super().__init__(*args, **kwargs)

File ~/Documents/MONAI-DEPLOY/monai-deploy-app-sdk/monai/deploy/core/application.py:126, in Application.__init__(self, runtime_env, do_run, path)
    123 self._builder()
    125 # Compose operators
--> 126 self.compose()
    128 if do_run:
    129     self.run(log_level=args.log_level)

Cell In [5], line 12, in AISpleenSegApp.compose(self)
     10 series_to_vol_op = DICOMSeriesToVolumeOperator()
     11 # Creates DICOM Seg writer with segment label name in a string list
---> 12 dicom_seg_writer = DICOMSegmentationWriterOperator(seg_labels=["Spleen"])
     14 # Creates the model specific segmentation operator
     15 spleen_seg_op = SpleenSegOperator()

TypeError: __init__() missing 1 required positional argument: 'segment_descriptions'

```

vikashg:
Hi @MMelQin, @gigony do we have the liver image somewhere on google drive that I can use to run some tests. Notebooks are self contained so I can run and test. I got an error, I think its a small one.

monai/deploy/operators/monai_bundle_inference_operator.py
```diff
@@ -477,14 +485,19 @@ def compute(self, op_input: InputContext, op_output: OutputContext, context: Exe
...
+                value = MetaTensor.ensure_torch_and_prune_meta(value, meta_data)
                 inputs[name] = value
-                if metadata:
-                    inputs[(f"{name}_{self._meta_key_postfix}")] = metadata
+                # Named metadata dict not needed any more, as it is in the MetaTensor
 
             inputs = self.pre_process(inputs)
-            first_input = inputs.pop(first_input_name)[None].to(self._device)  # select first input
-            input_metadata = inputs.get(f"{first_input_name}_{self._meta_key_postfix}", None)
+            first_input_v = inputs[first_input_name]  # keep a copy of value for later use
```
Nic-Ma, line 499: Do we need `deepcopy` here?

Thanks.

monai/deploy/operators/monai_bundle_inference_operator.py
```diff
@@ -709,18 +723,20 @@ def _convert_from_image_dicom_source(self, img: Image) -> Tuple[np.ndarray, Dict
                 img_meta_dict["depth_pixel_spacing"],
             ]
         )
-        meta_dict["original_affine"] = np.asarray(img_meta_dict.get("nifti_affine_transform", None))
-        meta_dict["affine"] = meta_dict["original_affine"]
+        # Use defines MetaKeys directly
```
Nic-Ma, line 726: I don't quite understand this line:
```py
meta_dict = {key: img_meta_dict[key] for key in img_meta_dict.keys()}
```
Maybe just `meta_dict = dict(img_meta_dict)`?

Thanks.

Nic-Ma (approved):

monai/deploy/operators/monai_bundle_inference_operator.py
MMelQin: Good catch. I think it was meant for deepcopy. Changed accordingly with copy.deepcopy to make it clear.

monai/deploy/operators/monai_bundle_inference_operator.py
MMelQin: The InvertD has changed to look for `image` dict name instead of `image_metadata_dict`. It was not clear to me if it matters the [`image`] value is before or post inference or even preprocessing. So, the intend here is really just to keep a reference to the value part of dict for adding back the key, `image`, because of the `pop` of key, `image`. I tried not to use `pop` but somehow direct using `inputs[first_input_name].to(self._device)` did not work.

wyli (approved):
Thanks, it looks good to me.

PR #366, opened 2022-10-09
Add DICOM PDF Writer and split dicom_utils common module out of Text SR Writer

MMelQin:
This PR is to:
- split from the existing text SR writer the dicom_utils module for reusable classes and function, e.g. equipment info, model info, DICOM common module writer for study, patient, SOP commons, etc.
- the DICOM Text Writer becomes much smaller and clearer to read
- the DICOM Encapsulated PDF write is added with minimal amount of code. This new writer take in either PDF bytes or a file path to load byte from, optionally with a DICOM Series for copying study level tags, and writes the DICOM Encapsulated PDF instance file. The loaded PDF in bytes are validated for being PDF with the use of PyPDF2  
- the app that uses the Text SR writer, mednist classification, is updated with simple change. This app is also used as the tester, even though the writers have test function for testing the operator standalone
- the license for the third party PyPDF2 is added

By the way, the SonarCloud failures are false positives, on some bogus code duplication findings. I have reviewed them and marking them as info.

Screenshot of the DICOM Text SR and Encapsulated PDF viewed in MicroDICOM:

image

image

Screenshot of actual PDF opened from MicroDICOM
image

   

MMelQin:
Reviewed off-line.

PR #368, opened 2022-10-13
Various fixes to PIN integration example

CPBridge:
Hi @aihsani 

Thanks again for this nice integration example and sorry for taking so long to get around to looking over it properly. I just played around with it for a while and it seems to be working quite nicely. During this process, I made a bunch of small changes, which I've collected in this pull request:

- A few fixes to the documentation (small typos and things)
- Despite the docs saying that a GPU is not required, the device for the inference part is actually hard coded as "cuda", meaning it won't work on CPU and will just crash. I fixed this. Unfortunately this does affect me as I cannot figure out how to get GPUs active within a Docker container when using the version of Docker compose that I have (if you have a quick pointer to a fix I'd appreciate it but I don't necessarily need a solution).
- I did make a fairly substantial change to the way the GSPS output works. Even though this is not the point of the integration, I think it's important that this follows best practices for GSPS creation. Generally speaking, a single GSPS should relate to an entire series, not one GSPS object per image or detection (as currently implemented). Generally speaking, if there are multiple GSPS objects for a series, a viewer (including eUnity) will only apply a single, manually-selected GSPS at a time. Therefore the current way things work would give rise to a pretty bad UI in the viewer. Therefore, I tweaked this to created a single GSPS for the entire series. This also has the effect of applying the windowing to the whole series.


aihsani (approved):

MMelQin:
> * Despite the docs saying that a GPU is not required, the device for the inference part is actually hard coded as "cuda", meaning it won't work on CPU and will just crash. I fixed this. Unfortunately this does affect me as I cannot figure out how to get GPUs active within a Docker container when using the version of Docker compose that I have (if you have a quick pointer to a fix I'd appreciate it but I don't necessarily need a solution).

This may not be exactly what you are seeking for, but to achieve this we use NV container base, and run docker with GPU options (`--gpus` or `--runtime=nvidia`), and for Docker Compose [by defining `driver` ](https://docs.docker.com/compose/gpu-support/)

integrations/nuance_pin/app/inference.py
```diff
@@ -181,7 +181,7 @@ def pre_process(self, img_reader) -> Compose:
                 ToTensord(
                     keys=image_key,
                 ),
-                ToDeviced(keys=image_key, device="cuda"),
+                ToDeviced(keys=image_key, device=self.device),
```
MMelQin, line 184: This comment is specific about this line, but Line 181.
Please rebase, as the inference.py has been updated in a previous PR by me to make the App SDK compositable with MONAI Core 0.9.1 and newer; for this one there is a small change in the pre-processing to remove the use of ToTensord

CPBridge:
> This may not be exactly what you are seeking for, but to achieve this we use NV container base, and run docker with GPU options (--gpus or --runtime=nvidia), and for Docker Compose by defining driver 

Thanks Min, but this hasn't been working for me. It's ok, no need to worry about his here.

> Please rebase, as the inference.py has been updated in a previous PR by me to make the App SDK compositable with MONAI Core 0.9.1 and newer; for this one there is a small change in the pre-processing to remove the use of ToTensord

Rebased, and fixed some style issues (that bizarrely were not caught before)

aihsani:
Ready to merge?

aihsani (approved):

MMelQin (commented):
LGTM

CPBridge:
@aihsani I made a couple of small tweaks to the README, as we discussed on slack

MMelQin (approved):
LGTM

PR #370, opened 2022-10-14
Enhanced multi-value DataElement parsing

MMelQin:
This PR enhances the code to treat multi-value DataElement retrieved from SOP instance Dataset differently from single valued ones.

DataElement retrieved from the underlying SOP instance Dataset may be multiple-valued; the retrieval is only needed when the to-be-matched attribute is not in the defined set of study and series attributes and when instance level attributes are used. The existing implementation favors multiple-valued and converts the string representation of the multi-valued into a list, however this also makes a single value into a list which then causes issue when valuating the condition.
   

vikashg (approved):
I do not see any problems with this code change. 
Looks good to me.

diazandr3s (approved):
LGTM

PR #372, opened 2022-10-16
Add the missing python package, PyPDF2

MMelQin:
The dependency introduced by the DICOM PDF Writer needs to be added to the requirements as well as in the Operator package list.
Also removed the spaces in a requirements file.


vikashg:
Hi @MMelQin, 
I am trying to test this using a TCIA dataset as I see in the test function in the dicom_encapsulated_pdf_writer_operator.py 
Does it need any TestPDF.pdf file. 
I am missing this file in the repo to complete the test.

MMelQin:
> Hi @MMelQin, I am trying to test this using a TCIA dataset as I see in the test function in the dicom_encapsulated_pdf_writer_operator.py Does it need any TestPDF.pdf file. I am missing this file in the repo to complete the test.

Sorry the test data/file are not going into the repo. I used MS Word to generated a TestPDF.pdf file.

vikashg:
Hi @MMelQin I do not have any issue with this merge everything works I tested. However, we must think of a proper example

MMelQin:
> Hi @MMelQin I do not have any issue with this merge everything works I tested. However, we must think of a proper example

Thanks @vikashg . Good idea to provide examples! I can think of a classification application where after the inference a custom operator is used to convert the results into PDF and textual report, and then both the DICOM Encapsulated PDF and Simple Test SR Writer are used to wrap the content into corresponding DICOM objects.

Currently we have the [MEDNIST example app](https://github.com/Project-MONAI/monai-deploy-app-sdk/tree/main/examples/apps/mednist_classifier_monaideploy) that uses the Text SR writer, and maybe it can be extended to include a text/html to PDF operator.

I have some code based on [headless-pdfkit](https://pypi.org/project/headless-pdfkit/), though it needs the wkhtmltopdf and xvfb-run wrapper. There is also the [fpdf ](https://pypi.org/project/fpdf/) which can directly convert text to PDF with a bit of formatting info. So, we'll leave it to the app dev to pick which one is more appropriate for the use case.

mingxin-zheng (approved):

vikashg (approved):
Its all good and working for me

PR #377, opened 2022-10-19
Merge Nuance Updates

iain-henderson:
Currently, the sample service has stopped being functional.

This pull request may be updated if I can root out the cause of that.


aihsani:
I merged a small change into main that allows the service to work again.

aihsani (approved):

PR #378, opened 2022-10-21
Enhanced code to extract and parse specific config files from TorchScript archive

MMelQin:
Per bug report, the Config Parser needs to parse the full list of configs all at once.
- As before, try to avoid extracting files to tmp dir, so if only one item is needed, directly read from the archive (into bytes). The helper function to extract content directly from a single file in the archive has been enhanced to try both supported file types, e.g. .json and .yaml.
- When list of config (files) are need to be parsed
  - considered using mem mapped files to hole content retrieved from the archive, however, the underlying config parser function, load_config_file, does not support mmap type
  - have to use tempfile.TemporaryDirectory as destination for extracting only the files for the named configs, both .json and .ymal will be tried, first match wins
  - Tested with sample app which was updated to illustrate the use of setting config_names even though currently only `inference` is needed, beside `metadata`

### DIFF
--- a/monai/deploy/operators/monai_bundle_inference_operator.py
+++ b/monai/deploy/operators/monai_bundle_inference_operator.py
@@ -63,7 +63,7 @@ def get_bundle_config(bundle_path, config_names):
     Gets the configuration parser from the specified Torchscript bundle file path.
     """
 
-    bundle_suffixes = [".json", ".yaml"]  # The only supported file ext(s)
+    bundle_suffixes = (".json", ".yaml", "yml")  # The only supported file ext(s)
     config_folder = "extra"
 
     def _read_from_archive(archive, root_name: str, config_name: str, do_search=True):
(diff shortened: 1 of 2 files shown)


monai/deploy/operators/monai_bundle_inference_operator.py
```diff
@@ -62,43 +63,110 @@ def get_bundle_config(bundle_path, config_names):
     Gets the configuration parser from the specified Torchscript bundle file path.
     """
 
-    def _read_from_archive(archive, root_name: str, relative_path: str, path_list: List[str]):
-        """A helper function for reading a file in an zip archive.
+    bundle_suffixes = [".json", ".yaml"]  # The only supported file ext(s)
```
Nic-Ma, line 66: I would suggest to refer to the MONAI ConfigParser for suffixes:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/config_parser.py#L381

Thanks.

Nic-Ma (approved):

monai/deploy/operators/monai_bundle_inference_operator.py
MMelQin: Addressed in the last commit.

PR #380, opened 2022-10-25
Fix app requirements to working version

aihsani:


MMelQin (approved):
LGTM. Pinning to the specific version as newer version of monai and App SDK are not compatible, root cause yet to be found.

PR #385, opened 2022-11-02
ENH: Versions as strings, not numbers, so that 3.1 != 3.10, etc.

Leengit:
Re-writes version numbers as strings rather than as numbers in `.github/workflows/*.yml` files because 3.1 != 3.10, etc.  Nothing is broken at present but in the future as, e.g., a Python version goes from 3.8 to 3.9 to 3.10, this could end up being an unexpected problem.

Signed-off-by: Lee Newberg <lee.newberg@kitware.com>


MMelQin (approved):
Thanks for catching this.

PR #387, opened 2022-11-03
Fix runtime error for MAP of breast density classification app

MMelQin:
Needed to provided this fix after building and testing MAP of the breast density classification app
- Somehow highdicom dependency is dragged in. Add this package in the App pip_packages decorator to avoid runtime error.
- Need to change the app.py __main__ function to directly init and run the app object, because the packager sets the command to run app.py instead of app/. Need to rename the original main() which served as test function to test()
- Published the MAP on ghcr.io, http://ghcr.io/mmelqin/monai_ai_breast_density_cls_app:1.0, to be used as a candidate in the AWS Imaging Service integration demo

### DIFF
--- a/examples/apps/breast_density_classifer_app/app.py
+++ b/examples/apps/breast_density_classifer_app/app.py
@@ -1,12 +1,13 @@
 from breast_density_classifier_operator import ClassifierOperator
 
-from monai.deploy.core import Application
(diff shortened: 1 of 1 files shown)


vikashg (approved):
Thanks @MMelQin I tested it works.

PR #394, opened 2022-11-12
Make the output DICOM SR instance part of the original study

MMelQin:
The Breast Density Classification example app generates DICOM SR instance as if in a new study. This by itself is supported if the original input DICOM is not available, e.g. in cases where input images are non-DICOM files.

The example as created expects to load and parse input DICOM files, so it's expected that the generated DICOM SR instance will copy the study level tags of the input, in order to be of the same study.


vikashg (approved):
Looks good to me.

PR #406, opened 2023-03-13
enable AMD GPU

vickytsang:


monai/deploy/packager/util.py
```diff
@@ -42,7 +42,11 @@ def verify_base_image(base_image: str) -> str:
         str: returns string identifier of the dockerfile template to build MAP
         if valid base image provided, returns empty string otherwise
     """
-    valid_prefixes = {"nvcr.io/nvidia/cuda": "ubuntu", "nvcr.io/nvidia/pytorch": "pytorch"}
+    import torch
+    if "AMD" not in torch.cuda.get_device_name(0):
```
MMelQin, line 46: This introduces a limitation on the Packager in that a GPU must be present on the local host, and its type determines if  what the built container is for AMD of not. This means,
1. Cannot build a container with AMD base on a host with Nvidia GPU
2. The script fails when none GPU is visible; this can be tested with setting Nvidia GPU to invisible. 
```
     import os
     os.environ['CUDA_VISIBLE_DEVICES'] = ''
     import torch
     torch.cuda.is_available()  # -> False
     torch.cuda.get_device_name(0)  # -> RuntimeError: No CUDA GPUs are available
```
    Also, `torch.cuda.get_device_name()` causes the script to hang when GPU not available 

So, at least a try/except block is needed to prevent the failure, and make known an MAP targeting AMD GPU cannot only be built on host with an AMD GPU.

monai/deploy/packager/util.py
```diff
@@ -89,12 +93,22 @@ def initialize_args(args: Namespace) -> Dict:
     if args.base:
         dockerfile_type = verify_base_image(args.base)
         if not dockerfile_type:
-            logger.error(
-                "Provided base image '{}' is not supported \n \
-                          Please provide a Cuda or Pytorch image from https://ngc.nvidia.com/ (nvcr.io/nvidia)".format(
-                    args.base
+            import torch
+            if "AMD" not in torch.cuda.get_device_name(0):
```
MMelQin, line 97: Same comment here as above.

monai/deploy/runner/runner.py
```diff
@@ -87,7 +87,9 @@ def run_app(map_name: str, input_path: Path, output_path: Path, app_info: dict,
     # Use nvidia-docker if GPU resources are requested
     requested_gpus = get_requested_gpus(pkg_info)
     if requested_gpus > 0:
-        cmd = "nvidia-docker run --rm -a STDERR"
+        import torch
+        if "AMD" not in torch.cuda.get_device_name(0):
```
MMelQin, line 91: Same comment here.

monai/deploy/runner/runner.py
```diff
@@ -160,6 +162,8 @@ def pkg_specific_dependency_verification(pkg_info: dict) -> bool:
     """
     requested_gpus = get_requested_gpus(pkg_info)
     if requested_gpus > 0:
+        import torch
+        if "AMD" not in torch.cuda.get_device_name(0):
```
MMelQin, line 166: Same here.

MMelQin:
Thank you @vickytsang for the pull request.

Do all AMD GPU device names contain the word "AMD"? I have no access to a AMD GPU, and it will be great if you can provide some reference. Also, for the same reason, I cannot test a built package targeting AMD GPU.

I have also left comments in the code, mostly on error handling.

monai/deploy/utils/deviceutil.py
```diff
@@ -0,0 +1,33 @@
...
+
+def has_rocm():
+    """Return True if ROCm is installed and GPU device is detected.
+
+    Args:
+
+    Returns:
+        True if ROCm is installed and GPU device is detected, otherwise False.
+    """
+    cmd = "rocminfo"
```
MMelQin, line 23: The command errors out when no ROCm installed. So the method should have a try/except block on calling the rocminfo command, and return false on except.

vickytsang: Do the new commit address these issues?

MMelQin (commented):
Comments have been addressed.

MMelQin:
We'll merge this PR as experimental support since
- this only affects packager on what additional base image can be supported
- the App SDK itself does not have explicit dependency on CUDA or GPU devices, rather, it is the (custom) operators and applications that may do, and the user shall ensure app works with ROCm and AMD GPU for packaging
- App SDK Packager separately has been going through some major changes and will come from a underlying dependency package in the next release of the App SDK.

PR #411, opened 2023-03-21
Fix issue 410 where the latest typeguard version 3.x is incompatible

MMelQin:
This PR is intended to only change the version of typeguard to ~=2.12.1 (which is equivalent "typeguard>=2.12.1, <3") in the requirements file as well as the Jupyter Notebooks, as it had been discovered that the newer v3 caused application runtime issues when validating the IO_context. Limiting the typeguard version to below 3 avoids the false error, and is used as the solution for the time being while the App SDK is migrated.

However, more files had to be updated, not related to the intended changes but more to resolve the complaints from the newer versions of Flake8, pytype, and mypy,
- the existing setup.cfg is not compatible with the latest pytype, v2023.03.13, with one symptom being the inputs could not be parsed out by pytype. So, new config was regenerated and added to pyproject.toml, as the newer pytype has added support for it
- Flake8 complained about `B905 `zip()` without an explicit `strict=` parameter,` but this is conflicting with pytype and mytype. So, this B905 is ignored in Flake8 settings
- pytype complained about `File "/home/mqin/src/monai-deploy-app-sdk/monai/deploy/core/application.py", line 222, in add_flow: New container type for io_maps does not match type annotation [container-type-mismatch]`, and this error is ignored in the pytype settings because the App SDK will stop using these class when migrated, and, there was no runtime errors
- pytype complained about `File "/home/mqin/src/monai-deploy-app-sdk/monai/deploy/operators/dicom_series_to_volume_operator.py", line 229, in prepare_series: No attribute 'first_pixel_on_slice_normal' on monai.deploy.core.domain.dicom_sop_instance.DICOMSOPInstance [attribute-error]`. This error is ignored for now, and a new [bug ](https://github.com/Project-MONAI/monai-deploy-app-sdk/issues/413) has been created for it

With the change and editable App SDK, verified that the newly built MONAI App Package (docker) indeed has the correctly typeguard version 2.12.1 (not 3.0.1 as reported in the issue)

Also tested the Jupyter notebook and verified the all steps executed successful (e.g. running as script, with `exec` command, with `package` and `run` command).

### DIFF
--- a/examples/apps/ai_livertumor_seg_app/livertumor_seg_operator.py
+++ b/examples/apps/ai_livertumor_seg_app/livertumor_seg_operator.py
@@ -52,14 +52,12 @@ class LiverTumorSegOperator(Operator):
     """
 
     def __init__(self):
-
         self.logger = logging.getLogger("{}.{}".format(__name__, type(self).__name__))
         super().__init__()
         self._input_dataset_key = "image"
         self._pred_dataset_key = "pred"
 
     def compute(self, op_input: InputContext, op_output: OutputContext, context: ExecutionContext):
-
         input_image = op_input.get("image")
         if not input_image:
             raise ValueError("Input image is not found.")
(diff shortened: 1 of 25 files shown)


MMelQin:
This PR's main change is to limit the typeguard version range, and the rest of the changes are related to formatting issues reported by the new versions of the format and type checking tools used in the build process. The changes were tested and reviewed outside of github, so the PR can be merged.

PR #428, opened 2023-05-19
Added option for SimpleInferer 

vikashg:
Added an option for SimpleInferer on monai_seg_inference_operator.py

### DIFF
--- a/examples/apps/ai_livertumor_seg_app/livertumor_seg_operator.py
+++ b/examples/apps/ai_livertumor_seg_app/livertumor_seg_operator.py
@@ -15,7 +15,7 @@ from numpy import uint8
 
 import monai.deploy.core as md
 from monai.deploy.core import DataPath, ExecutionContext, Image, InputContext, IOType, Operator, OutputContext
-from monai.deploy.operators.monai_seg_inference_operator import InMemImageReader, MonaiSegInferenceOperator
+from monai.deploy.operators.monai_seg_inference_operator import InfererType, InMemImageReader, MonaiSegInferenceOperator
 from monai.transforms import (
     Activationsd,
     AsDiscreted,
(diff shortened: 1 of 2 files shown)


vikashg:
@MMelQin Including `StrEnum` gives me an error for testing. If you can have a look that will be really helpful. Maybe I am using it wrong.

MMelQin:
> @MMelQin Including `StrEnum` gives me an error for testing. If you can have a look that will be really helpful. Maybe I am using it wrong.

Yes, needs to define the class of type StrEnum before using it. Will suggest some code.

vikashg:
Hi @MMelQin,
So I looked up `Enum` which is similar to `StrEnum`. There are existing examples like 
image

But I don't know how this will help or enhance the existing code. Please let me know and I will fix it. 
Thanks 
Vikash

MMelQin:
> Hi @MMelQin, So I looked up `Enum` which is similar to `StrEnum`. There are existing examples like image
> 
> But I don't know how this will help or enhance the existing code. Please let me know and I will fix it. Thanks Vikash

Another issue is that the MD App SDK requires Python version >= 3.7, for which the StrEnum is NOT available. This is why MONAI Core had [its impl of StrEnum](https://github.com/Project-MONAI/MONAI/blob/f355d1fc38c50e83e4f20ffe870c19a5814dc501/monai/utils/enums.py#L65), as Eric mentioned. For the upcoming version of the SDK, v0.6, I've changed to min version of Python >=3.8, but for 0.5 and any patch release of it, it is better to still stay with Python >=3.7. 

For our use case, the relevant magic strings had not been used in the App SDK or app code, I think we can safely use Enum or IntEnum for now.

MMelQin (commented):
LGTM

MMelQin (approved):

PR #432, opened 2023-06-05
Fix broken links in Readme

MMelQin:
Some links are broken due to the change of sample app name (removal of the numeral prefix).


MMelQin:
All links in the Readme were checked to be working, and no further review needed. It is noted that the links point to the `latest` version of the monai.io docs, though will be changed to `stable` version later.

PR #435, opened 2023-06-23
Updated the mednist notebooks adding the dependencies of pydicom and highdicom

MMelQin:
These packages are needed when packaging the app into app package container image. After the change, the notebooks ran successfully. Besides, all other notebooks have also been tested and passed.

Also needed to address a issue with PIL image where it erroneously interpret uint8 with values [0 | 1] as float.


examples/apps/simple_imaging_app/gaussian_operator.py
```diff
@@ -25,12 +25,20 @@ class GaussianOperator(Operator):
...
+        import numpy as np
         from skimage.filters import gaussian
         from skimage.io import imsave
 
         data_in = op_input.get().asnumpy()
         data_out = gaussian(data_in, sigma=0.2, channel_axis=2)  # Add the param introduced in 0.19.
 
+        # Make sure the data type is what PIL Image can support, as the imsave function calls PIL Image fromarray()
+        # Some details can be found at https://stackoverflow.com/questions/55319949/pil-typeerror-cannot-handle-this-data-type
+        print(f"Data type of output: {type(data_out)!r}, max = {np.max(data_out)!r}")
```
mocsharp, line 37: Any reason not to use logging fx?

mocsharp (approved):

examples/apps/simple_imaging_app/gaussian_operator.py
MMelQin: Good point, though this had been created as a simple example for illustration purpose only, not part of the SDK at all.

Issue #433, opened 2023-06-12
[BUG] OS error when building deploy app


filipmu:
**Describe the bug**

Get an `OSError: source code not available` when creating the deploy application.

```---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
Cell In[6], line 1
----> 1 app = App()

File /opt/conda/lib/python3.10/site-packages/monai/deploy/core/application.py:95, in Application.__init__(self, runtime_env, do_run, path)
     93     self.path = Path(path)
     94 else:
---> 95     self.path = get_class_file_path(self.__class__)
     97 # Set the runtime environment
     98 if str(self.path) == "ipython":

File /opt/conda/lib/python3.10/site-packages/monai/deploy/utils/importutil.py:116, in get_class_file_path(cls)
    100 """Get the file path of a class.
    101 
    102 If the file path is not available, it tries to see each frame information
   (...)
    112     A file path of the class.
    113 """
    115 try:
--> 116     return Path(inspect.getfile(cls))
    117 except TypeError:
    118     # If in IPython shell, use inspect.stack() to get the caller's file path
    119     stack = inspect.stack()

File /opt/conda/lib/python3.10/site-packages/torch/package/package_importer.py:691, in _patched_getfile(object)
    689     if object.__module__ in _package_imported_modules:
    690         return _package_imported_modules[object.__module__].__file__
--> 691 return _orig_getfile(object)

File /opt/conda/lib/python3.10/inspect.py:785, in getfile(object)
    783             return module.__file__
    784         if object.__module__ == '__main__':
--> 785             raise OSError('source code not available')
    786     raise TypeError('{!r} is a built-in class'.format(object))
    787 if ismethod(object):

OSError: source code not available
```

**Steps/Code to reproduce bug**

Run [this](https://gist.github.com/filipmu/d929a696a9b0dd4c4623467a15533019) notebook in jupyterlab.

**Expected behavior**

No stack trace, successful instantiation.

**Environment details (please complete the following information)**

- OS/Platform: `Linux jupyter-u63dd2b35b6fb5b--w646fbf22bc325c 5.15.0-1035-azure #42-Ubuntu SMP Tue Feb 28 19:41:23 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
- Python Version:Python 3.10.10
- Method of MONAI Deploy App SDK install: pip
- SDK Version:monai-deploy-app-sdk-0.5.0

**Additional context**

[...]

PR #436, opened 2023-06-26
Fixes Project-MONAI/monai-deploy-app-sdk#433

filipmu:
A fix to handle a change in python 3.10 for the inspect.getfile function. It now returns an `OSError: source code not available`


MMelQin (approved):
LGTM. Thanks for the contribution!

PR #438, opened 2023-07-11
Update SDK to support typeguard v3.0 and above, which has a breaking change in its API, and updated notebooks

MMelQin:
The key change for this PR is to

- Update the SDK to support typeguard v3.0 and above, which has a breaking change in its API

During testing, the following changes had to be made

- Update the SDK to require Python >= 3.8
- Update pytest version to latest version
- Update the run script to use the new way of calling pytest, `-m pytest` instead of `-m py.test`
- Update code to fix _black_ and _flake8_ complaints, and sometimes they have conflicts
- Test all of the notebooks, and the results can be used as QA test results (during testing not setting the `--upgrade` on importing monai-deploy-app-sdk, so the editable sdk is used, though for users the `--upgrade` so that they do not need to explicitly uninstall older version of the App SDK)

Code coverage only includes the unit tests, not the tests of the notebooks which provide additional and fuller coverage.


tests/unit/test_runner.py
```diff
@@ -168,8 +168,8 @@ def test_run_app_for_input_output_path_with_space(
     expected_container_output /= app_manifest["working-directory"]
 
     returncode = runner.run_app(sample_map_name, input_path, output_path, app_manifest, faux_pkg_manifest, quiet)
-    input_path_with_quotes = f'"{input_path.absolute()}"'
-    output_path_with_quotes = f'"{output_path.absolute()}"'
+    input_path_with_quotes = '"' + f"{input_path.absolute()}" + '"'  # To quiet flake8 complaint
+    output_path_with_quotes = '"' + f"{output_path.absolute()}" + '"'
```
gigony, line 172: Looks good (I am okay as it is).

Just wonder if

```
    input_path_with_quotes = f'''"{input_path.absolute()}"'''
    output_path_with_quotes = f'''"{output_path.absolute()}"'''
```
shows the same flake8 complaints.

```
    input_path_with_quotes = '"' + str(input_path.absolute()) + '"'  # To quiet flake8 complaint
    output_path_with_quotes = '"' + str(output_path.absolute()) + '"'
```
Above also can do the same.

gigony (approved):
Thanks @MMelQin !
It looks good to me!

cumo80 (approved):

mocsharp (approved):

tests/unit/test_runner.py
MMelQin: Yes, yet another way that works, in addition to str.format(), a leading or trailing space, but escape does not work as black would complain and remove it. Thx.

PR #439, opened 2023-07-13
Update Read The Docs configuration to support Python 3.8

gigony:
`stable` image doesn't support Python 3.8, causing the following issue.

```
Problem in your project's configuration. Invalid "python.version": expected one of (2, 2.7, 3, 3.5, 3.6, 3.7, pypy3.5), got 3.8
```

https://docs.readthedocs.io/en/stable/config-file/v2.html#build-image-legacy

image

And looks like we are using the old configuration.

This PR updates the configuration to use the latest syntax.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -25,4 +25,4 @@ python:
 
 
 build:
-  image: stable
+  image: latest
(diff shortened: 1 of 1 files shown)


MMelQin (approved):
Thanks Gigon for finding out the issue!

PR #471, opened 2023-11-16
Address a removed monai function, fix CUDA 11 runtime lib, and resurrect option to omit or not DICOM SEG empty frame

MMelQin:
This PR expanded from its initial scope to include addition changes required.
- first is to update the code where the MONAI Core transforms function `AddChannel` is imported and used, namely in the MedNIST and Nuance PIN integration example applications. MONAI Core 1.3 was release in mid Oct, 2003, and the aforementioned function was removed, and `EnsureChannelFirst` function is to be used. Most examples had already been updated to make use the new function, except the aforementioned ones.
- second is that the in PR triggered Github CI actions, pytype check failed due to failure to find libcudart.so.11.0 when importing holoscan. Limiting torch to v2.0.1 to piggyback on ii to find CUDA runtime lib did not work, and had to explicitly export LD_LIBRARY_PATH pointing to the Python version specific site-packages folder.
- third is that the DICOM Seg Writer in App SDK v0.6 somehow omitted the option of include empty frame. Added it back in, but since it is an built-in operator in the App SDK package, won't be effective until the next release of the SDK.


MMelQin:
Did the final changes and review with @gigony together, so a separate review in not needed.

PR #481, opened 2024-04-03
Updating the SDK to use newer and updated dependencies, namely Holoscan SDK

MMelQin:
As the title says, this PR is to update the App SDK to make use of the Holoscan SDK v1.0.3.
Updates are mostly around documentation and the example apps.

Testing was done on Ubuntu 22.04 LTX with CUDA 12.4, and due to the CUDA 12.4 compatibility, Linux Kernel 6.5 HWE is used. Both Python 3.8 and Python 3.10 env were used in testing.   
- all examples tested successfully
- all notebooks tested successfully
- documentation updated and release note for v1.0.0 added
- building documentation needs to be done in Python 3.8 environment as of now

Note:
Jupyter notebooks cannot run as is. This is because the MAP Packager pull the monai-deploy-app-sdk from Pypi unless command option is used to install the monai-deploy-app-sdk package from (the to be released) local wheel file. The example applications would work perfectly in the dev branch with sdk package set up in edit mode.


CONTRIBUTING.md
dbericat, line 1: Let's also point to general MONAI contributing guidelines

dbericat (approved):
LGTM

mocsharp (approved):

CONTRIBUTING.md
MMelQin: That is not an umbrella contributing guide as one might think, rather specific to MONAI Core with some overlapping general items.

In any case, added the reference while mentioning general as well as specifics.

Issue #510, opened 2025-01-03
[BUG] dicom_seg_writer_operator.py -> process_images(...) does not accept numpy array


kavmar:
Hi,

I would like to result my app a DICOM SEG and I am trying to use dicom_seg_writer for it.  This is how I emit the data from my segmentation operator:

emit_data = data[0]["pred"].cpu().numpy()
emit_data = np.squeeze(emit_data, 0)
emit_data = emit_data.T.astype(np.uint16)
             
op_output.emit(emit_data, self.output_name_seg)

I checked that the data is a numpy array in debugger. However, according to dicom_seg_writer_operator.py lines 291 a 292:

elif not isinstance(image, np.ndarray):
      raise ValueError("'image' is not a numpy array, Image object, or supported image file.")

While lines before are trying to convert image to numpy the 291 will not pass even if the image is numpy. 

Am I missing something?

@CPBridge as discussed in highdicom, you contributed this operator, could you please doublecheck?

Thanks

CPBridge:
Hi @kavmar I think the problem here is not exactly what you think that it is. 

It is true that highdicom requires that the segment *descriptions* it is passed to be consecutive, starting at 1. However in the `DICOMSegmentationWriterOperator` this should be handled already by [this line](https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/9608cd77d00a71089526cc8f46f17826ef99463c/monai/deploy/operators/dicom_seg_writer_operator.py#L216) which automatically numbers the segments for you according to their index in the list you pass.

However highdicom *does not* require that every described segment be present. It is perfectly acceptable for you to specify that "tumor" is class 2, and then pass an array that contains no 2s. In fact, this is very important because then the receiver can understand that there is no tumor present (as opposed to the presence of tumor not being assessed).

[...]

PR #515, opened 2025-01-15
DICOM Seg Writer operater: Fix for case where input image is already a numpy array

CPBridge:
Potential fix for #510 (@kavmar)


CPBridge:
Not sure why I'm getting errors on this that are unrelated to my changes

MMelQin (approved):
Thanks for fixing this. Rebased to get rid of the CI errors.
At one time there was a design decision to encourage users to use the SDK's own Image object (numpy array + metadata) so that the metadata can be validated along the way before being passed down to the seg writer. If numpy array is directly used, the user then needs to ensure that it represents a image compatible with the original input image.

PR #516, opened 2025-01-17
Update example apps' test data path, patch hsdk in local env, and ensure correct MD SDK is used in dev env

MMelQin:
This PR intends to fix multiple issues reported lately.
- When testing in dev env with CLI packager, holoscan was not installed in MAP. The root cause is NOT the packager. The local monai-deploy-app-sdk editable package's PKG-INFO has the version of 0.5.1+24.gc60696a.dirty, obviously wrong as it should be 2.0.0+xxx, as and as such, the CLI reads the version and pip installs monai-deploy-app-sdk 0.5.1 from pypi.org. The reason for setting the wrong version of monai deploy app sdk is because that versioneer is used, and it calls _version.py, which in turn uses Git to query the Github repo for the latest version. Somehow, Github repo version was not tagged/set correctly since 0.5.1.
Changed to use the _version.py generated on bumping version
- The second issue is in holoscan sdk v2.8 (in v2.7 also), with the missing MultiMessageConditonInfo in import. The fix will be in v2.9. A post install action has been added to the setup.py, to patch locally installed holoscan sdk
- Until a new MONAI Deploy App SDK package is released, the aforementioned patching action would not be available when a user installs the MD App SDK from pypi.org. To workaround the issue, the user can run the same script to patch the installed holoscan package, but packaging the app with CLI Packager will not have the holoscan patched in the MAP image, hence, adding the `holoscan==2.6.0` to the apps' requirements.txt file is the workaround
- Security policy changes have prevented sharing the application test data file anonymously, and the hosted files have changed locations. Hence the examples have been updated with new instructions and URLs


setup.py
```diff
@@ -9,21 +9,84 @@
...
+                pass
+
+            return False
+
+        if not needed_to_patch():
+            return
+
+        print("Patching holoscan as needed...")
+        spec = importlib.util.find_spec("holoscan")
+        if spec:
```
mocsharp, line 58: Do we need to warn if the module cannot be found somehow? Highly unlikely I guess?

mocsharp (approved):

setup.py
MMelQin: Since holoscan is a required dependency of monai-deploy-app-sdk, I'd think pip would have failed the install if holoscan is not present (well, the holoscan package could be uninstalled afterwards, but then it is post post-install actions)

PR #529, opened 2025-03-03
Fix DICOMSeriesToVolumeOperator casting bug

WillButAgain:
The array was being read in as int16 which caused some values to be truncated, it should instead be uint16. I also added a check that uint16 is sufficient after the slope/intercept transformation is applied. if it is not, it will use int32 instead.

### DIFF
--- a/monai/deploy/operators/dicom_series_to_volume_operator.py
+++ b/monai/deploy/operators/dicom_series_to_volume_operator.py
@@ -112,7 +112,7 @@ class DICOMSeriesToVolumeOperator(Operator):
         # with the NumPy array returned from the ITK GetArrayViewFromImage on the image
         # loaded from the same DICOM series.
         vol_data = np.stack([s.get_pixel_array() for s in slices], axis=0)
-        vol_data = vol_data.astype(np.int16)
+        vol_data = vol_data.astype(np.uint16)
 
         # For now we support monochrome image only, for which DICOM Photometric Interpretation
         # (0028,0004) has defined terms, MONOCHROME1 and MONOCHROME2, with the former being:
@@ -156,9 +156,14 @@ class DICOMSeriesToVolumeOperator(Operator):
 
         if slope != 1:
             vol_data = slope * vol_data.astype(np.float64)
-            vol_data = vol_data.astype(np.int16)
-        vol_data += np.int16(intercept)
-        return np.array(vol_data, dtype=np.int16)
+        vol_data += intercept
+
+        # Check if vol_data can be cast to uint16 without data loss
+        if np.can_cast(vol_data, np.uint16, casting='safe'):
+            vol_data = np.array(vol_data, dtype=np.uint16)
+        else:
+            vol_data = np.array(vol_data, dtype=np.int32)
+        return vol_data
 
     def create_volumetric_image(self, vox_data, metadata):
         """Creates an instance of 3D image.
--- a/monai/deploy/operators/dicom_series_to_volume_operator.py
+++ b/monai/deploy/operators/dicom_series_to_volume_operator.py
@@ -155,17 +155,28 @@ class DICOMSeriesToVolumeOperator(Operator):
         except KeyError:
             slope = 1
 
-        if slope != 1:
-            vol_data = slope * vol_data.astype(np.float64)
-        vol_data += intercept
 
-        # Check if vol_data can be cast to uint16 without data loss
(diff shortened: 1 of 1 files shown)


monai/deploy/operators/dicom_series_to_volume_operator.py
```diff
@@ -112,7 +112,7 @@ def generate_voxel_data(self, series):
         # with the NumPy array returned from the ITK GetArrayViewFromImage on the image
         # loaded from the same DICOM series.
         vol_data = np.stack([s.get_pixel_array() for s in slices], axis=0)
-        vol_data = vol_data.astype(np.int16)
+        vol_data = vol_data.astype(np.uint16)
```
MMelQin, line 115: This can only be safe done when the pixel data is in fact unsigned, and the following DICOM tag tells the fact (when value == 1)

Tag	(0028,0103)
Type	Required (1)
Keyword	PixelRepresentation
Value Multiplicity	1
Value Representation	Unsigned Short (US)

Example Values	
  1
  0

Enumerated Values:
0000H
    unsigned integer.
0001H
    2's complement

monai/deploy/operators/dicom_series_to_volume_operator.py
```diff
@@ -156,9 +156,14 @@ def generate_voxel_data(self, series):
...
-            vol_data = vol_data.astype(np.int16)
-        vol_data += np.int16(intercept)
-        return np.array(vol_data, dtype=np.int16)
+        vol_data += intercept
+
+        # Check if vol_data can be cast to uint16 without data loss
+        if np.can_cast(vol_data, np.uint16, casting='safe'):
+            vol_data = np.array(vol_data, dtype=np.uint16)
+        else:
+            vol_data = np.array(vol_data, dtype=np.int32)
```
MMelQin, line 165: I would not use np.int32 here. The slope and intercept both are decimals, so the vol_data in this block could well be  np.float32. Notice also line 158 casted the vol_data to np.float64 (but the original code casts it back to int16 erroneously), as there could be float64 pixel values, so the can_cast check (to float32) will be a safe way.

WillButAgain: This is something i was wondering about... whether the pixel array values should always be ints. I simply guessed that all pixel array values would be ints, so any slope/intercept scaling returning floats should be rounded to ints.

monai/deploy/operators/dicom_series_to_volume_operator.py
```diff
@@ -112,7 +112,8 @@ def generate_voxel_data(self, series):
         # with the NumPy array returned from the ITK GetArrayViewFromImage on the image
         # loaded from the same DICOM series.
         vol_data = np.stack([s.get_pixel_array() for s in slices], axis=0)
-        vol_data = vol_data.astype(np.int16)
+        if slices[0][0x0028,0x0103].value == 1:
```
MMelQin, line 115: value 0 is for unsigned integer.

monai/deploy/operators/dicom_series_to_volume_operator.py
```diff
@@ -156,9 +157,16 @@ def generate_voxel_data(self, series):
...
-        return np.array(vol_data, dtype=np.int16)
+        vol_data += intercept
+
+        # Check if vol_data can be cast to uint16 without data loss
+        if np.can_cast(vol_data, np.uint16, casting='safe'):
+            vol_data = np.array(vol_data, dtype=np.uint16)
+        elif np.can_cast(vol_data, np.float32, casting='safe'):
+            vol_data = np.array(vol_data, dtype=np.float32)
+        elif np.can_cast(vol_data, np.float64, casting='safe'):
+            vol_data = np.array(vol_data, dtype=np.float32)
```
MMelQin, line 168: Typo for dtype -> np.float64.

WillButAgain:
There is one more commit with additional changes. 

 The problem is we need the datatypes for the slope and intercept to be compatible with the volume data. uint16 and float32 are not compatible. We ended up needing to check the datatypes of all 3 (vol_data_, slope, intercept) in order to do any casting. Our solution is not elegant, but it will always find the lowest overhead datatype between the options of uint16, float32, and float64. 

There may be some benefit in checking again to see if the datatype can be further reduced after the slope/intercept transform is applied. We are open to discussion on this and could add that functionality as well if desired.

PR #533, opened 2025-04-07
DICOM Seg Writer Op Tag Writing Updates

bluna301:
From Contributing Equipment Sequence for DICOM SEG Writer [discussion](https://github.com/Project-MONAI/monai-deploy-app-sdk/discussions/528). `DICOMSegmentationWriterOperator` updates to align with the other DICOM writer operators and the [IHE Radiology AI Results](https://www.ihe.net/uploadedFiles/Documents/Radiology/IHE_RAD_Suppl_AIR.pdf) criteria.

This initial commit adds `model_info` as an optional parameter to the `DICOMSegmentationWriterOperator`, allowing for the `ContributingEquipmentSequence` nesting in the DICOM SEG output:

```python
# model info is algorithm information
my_model_info = ModelInfo(
    creator="CCHMC CAIIR", # institution name
    name=_algorithm_name, # algorithm name
    version=_algorithm_version, # algorithm version
    uid="0.0.1", # MAP version
)
```

image

@MMelQin, a few thoughts for you after this initial commit:

- Should we provide any checks to ensure that the algorithm name and version information that would be present in both the `segment_descriptions` and `model_info` arguments are equivalent? As currently implemented, users can input different values for each field
- Thoughts on enforcing `model_info` as a required or optional argument? I believe this is required for other DICOM writers
- I didn't add an `equipment_info` argument to Seg Writer as 3/4 of the tag values are currently hard coded into the operator (see below). If you think these tags should always contain the MONAI Deploy App SDK information, we can leave this how it is; if you think these tags should be dynamic based on user input, I can add `equipment_info` as an optional argument to the Seg Writer and pass these values accordingly 

```python
seg = hd.seg.Segmentation(
    source_images=dicom_dataset_list,
    pixel_array=image,
    segmentation_type=hd.seg.SegmentationTypeValues.BINARY,
    segment_descriptions=self._seg_descs,
    series_instance_uid=hd.UID(),
    series_number=random_with_n_digits(4),
    sop_instance_uid=seg_sop_instance_uid,
    instance_number=1,
    manufacturer="The MONAI Consortium", # hard coded
    manufacturer_model_name="MONAI Deploy App SDK", # hard coded
    software_versions=version_str, # hard coded to MONAI Deploy App SDK version
    device_serial_number="0000",
    omit_empty_frames=self._omit_empty_frames,
)
```

### DIFF
--- a/monai/deploy/operators/dicom_seg_writer_operator.py
+++ b/monai/deploy/operators/dicom_seg_writer_operator.py
@@ -1,4 +1,4 @@
-# Copyright 2021-2023 MONAI Consortium
+# Copyright 2021-2025 MONAI Consortium
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -27,6 +27,7 @@ generate_uid, _ = optional_import("pydicom.uid", name="generate_uid")
 ImplicitVRLittleEndian, _ = optional_import("pydicom.uid", name="ImplicitVRLittleEndian")
 Dataset, _ = optional_import("pydicom.dataset", name="Dataset")
 FileDataset, _ = optional_import("pydicom.dataset", name="FileDataset")
+PyDicomSequence, _ = optional_import("pydicom.sequence", name="Sequence")
 sitk, _ = optional_import("SimpleITK")
 codes, _ = optional_import("pydicom.sr.codedict", name="codes")
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


MMelQin (approved):
Looks good to me. Thanks for making the AI Results in the generated DICOM instances consistent.
I think it is OK to leave the Equipment info as is.  The the existing implementation already tries to parse out the App SDK version, so the software that generates the DICOM instance is correct. More important is the AI model info encapsulated in the contributing equipment, which is added by this PR.

bluna301:
Thanks Ming. I agree with leaving `EquipmentInfo` to be handled by the operator as is.

Maybe we should consider making `ModelInfo` a required argument at some point, but I think this is good for now!

PR #535, opened 2025-04-16
Duplicate Input DICOM Spatial Coordinates

bluna301:
Will flesh out comment more tomorrow, just wanted to get some preliminary notes down prior to tomorrow's meeting.

Solution stores each SOP instance in a new dictionary (not ideal) and checks if the combination of `StudyInstanceUID`, `SeriesInstanceUID`, and `ImagePositionPatient` has been seen before; if so, it removes the duplicate, with preference for the higher `AcquisitionNumber`. In cases where the `ImagePositionPatient` tag is not present (X-Ray images, some non-image DICOM files like SR, KO, PR, etc. modalities), no duplication check takes place. This approach is effective for handling the case discussed:

```bash
[2025-04-15 23:07:39,429] [INFO] (dicom_data_loader_operator.DICOMDataLoaderOperator) - No or invalid input path from the optional input port: None
[2025-04-15 23:07:39,816] [INFO] (dicom_data_loader_operator.DICOMDataLoaderOperator) - DICOM Images to Load: 312
[2025-04-15 23:07:39,838] [INFO] (dicom_data_loader_operator.DICOMDataLoaderOperator) - Duplicate spatial coordinates detected; removing duplicate SOP at IPP (-250.0, -250.0, -746.25) in Series 1.2.840.113619.2.405.3.185216268.497.1730719949.233; kept SOP instance with higher AcquisitionNumber (3 > 2)
[2025-04-15 23:07:39,845] [INFO] (dicom_data_loader_operator.DICOMDataLoaderOperator) - DICOM Images Loaded: 311
```
I tested on ~20 clinical studies each for CT, CR, and MR modalities:
- CT: removes duplicate when present (just on the 1 study); all other cases have no duplicates
- CR: no duplicates present (`ImagePositionPatient` is absent), therefore no changes from existing loading functionality
- MR: removes a significant number of duplicates per study (sometimes around ~50% of total study images); repeated `ImagePositionPatient` values seem to be quite common for T1 mDIXON, mDIXON-Quant, SE-EPI MRE series; will provide anonymized version to illustrate

---------

I wonder if it is worth having an input parameter to the operator to control whether or not duplicates are checked for. For example, if a user is targeting MR mDIXON series for inference (not sure how common this is), the duplicate check should probably be turned off to prevent a large number of SOP instances from not being loaded. 

Finally, I am wondering how arbitrary referencing the higher `AcquisitionNumber` value is; besides the single test case, do we have any additional justification for using this approach?

### DIFF
--- a/monai/deploy/operators/dicom_data_loader_operator.py
+++ b/monai/deploy/operators/dicom_data_loader_operator.py
@@ -1,4 +1,4 @@
-# Copyright 2021-2023 MONAI Consortium
+# Copyright 2021-2025 MONAI Consortium
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -12,7 +12,9 @@
 import logging
 import os
 from pathlib import Path
-from typing import List
+from typing import Dict, List, Optional, Tuple, cast
+
+from pydicom.dataset import Dataset
 
 from monai.deploy.core import ConditionType, Fragment, Operator, OperatorSpec
 from monai.deploy.core.domain.dicom_series import DICOMSeries
@@ -163,6 +165,7 @@ class DICOMDataLoaderOperator(Operator):
         study_dict = {}
         series_dict = {}
         sop_instances = []
+        sop_map: Dict[Tuple[str, str], Dict[Optional[Tuple[float, float, float]], List[Dataset]]] = {}
 
         for file in files:
             try:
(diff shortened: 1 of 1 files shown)


monai/deploy/operators/dicom_data_loader_operator.py
```diff
@@ -171,34 +174,94 @@ def _load_data(self, files: List[str]):
...
+
+                series_instance_uid = sop_instance[0x0020, 0x000E].value.name  # name is the UID as str
+
+                if series_instance_uid not in series_dict:
+                    series = DICOMSeries(series_instance_uid)
+                    series_dict[series_instance_uid] = series
+                    self.populate_series_attributes(series, sop_instance)
+                    study_dict[study_instance_uid].add_series(series)
+
+                # Prepare sop_map entry
```
MMelQin, line 205: Would it safer to move line 205 to 207 to the conditional block above? Understand the sop_map for a series only needs to be initialized when an instance of this series is first added.
Also, per DICOM standards, series instance UID is globally unique, so are study and SOP instance UIDs. Making a key with both Study and series instance UID is not really necessary (Study Id and Series Id are never unique, though)

monai/deploy/operators/dicom_data_loader_operator.py
```diff
@@ -171,34 +174,94 @@ def _load_data(self, files: List[str]):
...
+                            f"({exist_acq_num} > {curr_acq_num})"
+                        )
+                    elif curr_acq_num == -1:
+                        # AcquisitionNumber tag is absent for compared SOP instances - don't store current SOP instance
+                        self._logger.info(
+                            f"Duplicate spatial coordinates detected; removing duplicate SOP at IPP {ipp_tuple} "
+                            f"in Series {series_instance_uid}; AcquisitionNumber tags are absent"
+                        )
+                    else:
+                        # AcquisitionNumber tag values are equal for compared SOP instances - don't store current SOP instance
```
MMelQin, line 249: Wonder if we can compare the acquisition time and take the newer one (e.g. assuming the modality is smart and can re-acquire a new image at the same position when it detects the raw data is not good for reconstruction). Acquisition time is not mandatory and can be absent though.

The above would be the edge case of the edge cases, so the implementation as is should suffice.

MMelQin (approved):
Look good to me. Thanks a lot.

vikashg (approved):

monai/deploy/operators/dicom_data_loader_operator.py
bluna301: Good point - the reason I made this key was because I figured there could be cases where the same IPP value could appear in multiple different series. For example, if there are 3 axial series in a study, I don't think it would be impossible (although it's probably unlikely) that the same IPP value could be in all three series. If we didn't link the IPP value with the Study-Series key, we would potentially be erroneously removing these SOP instances; linking to a Study-Series key allows for us to investigate duplicates within a single series.

bluna301:
Thanks Ming. Future work may need to be done to make this functionality optional (say, for example, the desired series for inference is an mDIXON MR series that has duplicate `ImagePositionPatient` values). But I think this is good for now!

PR #536, opened 2025-04-19
Support remote inference on Triton Inference Server with ease of use

### LINKED ISSUE #212: [FEA] Add remote inference support with the use of Triton Inference Server
**Is your feature request related to a problem? Please describe.**
App SDK currently supports inference within the application process itself. This is simple and efficient for some use cases, though when multiple applications/models are hosted in the "production" environment, remote inference service, e.g. Triton, may be needed, so that the heavy use of resource of inference itself can be centrally managed.    
**Describe the solution you'd like**
Add remote inference support to the built-in inference operators in the App SDK, with runtime options, e.g. using strategy pattern, to support the choice of in proc, or remote inference.
**Describe alternatives you've considered**
Treating and managing the App, AKA MAP, as a whole.
**Additional context**

### LINKED ISSUE #458: [IMP] Setting up MONAI Model Server
**Is your enhancement request related to a problem? Please describe.**
Over the last six months, we have been building towards packaging the MONAI model in different flavors. The focus of these efforts is towards model sharing and 
1. Packaged as a MONAI Bundle and shared as a model zoo. 
2. We can now make Pythonic calls to a monai bundle for further fine-tuning or inferencing.
The MONAI bundles can also be used for creating MONAI Application Packages (MAPs), which are their own entities directed for inferencing. However, it should be pointed out that MONAI Application Packages do not need to run a MONAI Bundle. MAPs do support DICOM data (dicom I/O) loaders which is not the case for MONAI Bundles (at least natively).

The core focus for MONAI 1.3 release has been giving access to MONAI bundles through python function calls. 

We can now standardize the model training and inferencing using the MONAI bundle

In essence, there are various flavors for models to be shared and re-used. However, one key use case that we are missing is that we do not have a mechanism to set up a model server, where we can have model-store (a directory) where models (monai bundles) are dropped and inference can be made by making a call to the model store using a model name. 

**Describe the solution you'd like**
An ideal solution would be to set up a model-server where models can be dropped and inference can be made by calling the model using the model_name

**Describe alternatives you've considered**
An alternative would be to use torchserve and the .mar format (model archive). Pytorch provides a CLI `torch-model-archive` to generate a .mar file. These .mar files contain a torchscript model along with python handlers for data pre and post processing. The models can be invoked using a RestAPI (`curl`) in conjunction with the model name, and the IP address of the host computer.

However, torchserve only supports Rest API (AFAIK) which might be good for small images but maybe not a great idea for large dicom images and files. As the model inference will fail if the connection drops in between.
An alternative to the Rest API might be tus-py https://github.com/tus/tus-py-client.

We can try to write a thin wrapper on top of our MONAI Bundle so that we can utilize the `torchserve` model server or repackage our monai bundles as .mar file so that they can be used in torchserve. For medical imaging applications, where the dataset size is quite large even for a single patient, we can set up a data store with unique identifiers (UID) for each patient. Instead of sending the data using a restful call, we can pass the unique identifier and the model server can pull the appropriate data from the datastore and complete the inference. 

**Additional context**
I think continuing with the idea that there is a theme with each release. For 1.3 the main theme was pythonic support for MONAI Bundles. Maybe for 1.4 the main theme would be a model server for monai bundles. 

**Unclear to me**
One of the aspects of torchserve that is unclear to me is if the models in model server are loaded in the GPUs (i.e., if the models are hot) and ready to serve. If we can achieve this it will be really great.

MMelQin:
This pull request adds support for performing inference on remote Triton Inference Server, with ease of use
- updated the TritonModel class to fully support and encapsulate the Triton client of the specific model hosted in Triton Inference Server. This is done by parsing the key elements of Triton model config.pbtxt file (actual model files not required), and dynamically setting the input and output parameters using the parsed model configurations. User app is relived from setting any arguments except for the server network location, and in future versions, TLS keys for secure communication.
- enhanced the NamedModel to include the support of Triton model repository (which may have multiple folders each for a specific model), making it possible to support hybrid inference scenario, i.e. local in-proc hosted as well as Triton hosted models, all transparent to the user app.
- updated the app context class to support the setting for Triton server network location, as well as calling the instantiated TritonModel to connect to the remote server.
- Added and tested example app showcasing remote inference on Triton Inference Server.

Also added the following files to the example application folder:
- Example Triton Model repo for the client side, i.e. with only model folder and its config.pbtxt file.
- Example Shell script to set the required env vars, though if these are not set, the command line option can be used instead.

The Quality Gate failure is due to the example app has duplicated code, which is true as it has exactly the same app inference operator as the Spleen Seg app with in-proc Torch model hosting. We have each example app folder contain all necessary code for ease of use, hence have to tolerate the duplicates with a few hundred lines of code.

### DIFF
--- a/examples/apps/ai_remote_infer_app/__main__.py
+++ b/examples/apps/ai_remote_infer_app/__main__.py
@@ -1,4 +1,4 @@
-# Copyright 2021-2025 MONAI Consortium
+# Copyright 2025 MONAI Consortium
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/examples/apps/ai_remote_infer_app/__main__.py
+++ b/examples/apps/ai_remote_infer_app/__main__.py
@@ -1,0 +1,15 @@
+# Copyright 2021-2025 MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from app import AIRemoteInferSpleenSegApp
+
+if __name__ == "__main__":
+    AIRemoteInferSpleenSegApp().run()
--- a/examples/apps/ai_remote_infer_app/app.py
+++ b/examples/apps/ai_remote_infer_app/app.py
@@ -1,4 +1,4 @@
-# Copyright 2021-2025 MONAI Consortium
+# Copyright 2025 MONAI Consortium
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/examples/apps/ai_remote_infer_app/app.py
+++ b/examples/apps/ai_remote_infer_app/app.py
@@ -1,0 +1,143 @@
+# Copyright 2021-2025 MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import logging
+from pathlib import Path
+
+from pydicom.sr.codedict import codes  # Required for setting SegmentDescription attributes.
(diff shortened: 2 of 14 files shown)


examples/apps/ai_remote_infer_app/__main__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021-2025 MONAI Consortium
```
mocsharp, line 1: Should we set the copyright year to 2025 for the new files?

examples/apps/ai_remote_infer_app/app.py
```diff
@@ -0,0 +1,143 @@
+# Copyright 2021-2025 MONAI Consortium
```
mocsharp, line 1: Should we set the copyright year to 2025 for the new files?

examples/apps/ai_remote_infer_app/env_settings_example.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+export HOLOSCAN_INPUT_PATH="inputs/spleen_ct_tcia"
```
mocsharp, line 2: Do we need copyright here?

mocsharp (approved):
LGTM. Thank you.

examples/apps/ai_remote_infer_app/__main__.py
MMelQin: Good point. Was debating to have year range or just the year when the new file is introduced. Will update to have just the year when introduced.

Issue #544, opened 2025-07-05
How is model.ts generated from model-zoo


liqunfu:
MONAI bundle from model zoo can be exported:
python -m monai.bundle trt_export --net_id network_def --filepath wholeBody_ct_segmentation/models/model_trt.ts --ckpt_file wholeBody_ct_segmentation/models/model.pt --meta_file wholeBody_ct_segmentation/configs/metadata.json --config_file wholeBody_ct_segmentation/configs/inference.json --precision fp32 --use_trace "True"

In this case weights are stored in a constants folder in model_trt.ts.

However, deploy sdk expects weights in a data folder: https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/2ce415fbe61f2a486faf409f00ce61efbbff8ded/monai/deploy/core/models/torch_model.py#L103

this means the exported model_trt.ts is not usable with deploy sdk.

So my question is:
Is there a process to export bundles from model-zoo to torch script models that can be used in the monai deploy sdk?

MMelQin:
TorchScript itself does not save constants in a constants folder in the archive, see [here](https://github.com/pytorch/pytorch/blob/main/torch/csrc/jit/docs/serialization.md), and the model file mentioned in the issue was exported with --trt_export flag, so it is the MONAI Bundle that uses the constants folder instead of constants.pkl for TRT optimized model. TorchScript does mention about an optional tensors folder.

PR #546, opened 2025-07-21
Improve TorchScript model identification logic in torch_model.py

gigony:
- Updated copyright notice to reflect the years 2021-2025.
- Enhanced the logic for identifying TorchScript models by clarifying the required files and directories in the zip archive format.

Address #544


MMelQin (approved):
Thanks @gigony for addressing this!

Edit after the merge: TorchScript itself does not save constants in a constants folder in the archive, see [here](https://github.com/pytorch/pytorch/blob/main/torch/csrc/jit/docs/serialization.md), and the model file mentioned in the issue was exported with --trt_export flag, so it is the MONAI Bundle that added the constants folder instead of constants.pkl for trt optimized model.

PR #547, opened 2025-07-28
MONAISegInferenceOperator Additional Arguments

bluna301:
When translating the [MONAI CT TotalSegmentator MONAI Bundle](https://github.com/Project-MONAI/model-zoo/tree/dev/models/wholeBody_ct_segmentation) to a MAP, it was determined that the segmentations produced by the MONAI Bundle and MAP were not exactly aligned (i.e. DICE != 1 for all organs).

This seems to be due to the `MONAISegInferenceOperator` not accepting all the possible arguments that are accepted by the [sliding_window_inference](https://github.com/Project-MONAI/MONAI/blob/d388d1c6fec8cb3a0eebee5b5a0b9776ca59ca83/monai/inferers/utils.py#L42) method, specifically `mode` and `padding_mode` for this Bundle. Once a custom operator that accepted these input arguments was implemented into the MAP pipeline, the DICE for all organs was ~=0.99 when comparing to the Bundle.

This initial commit includes the logic for accepting and implementing `mode` and `padding_mode` arguments. Happy to discuss further if the additional missing `sliding_window_inference` arguments should be added as well.

### DIFF
--- a/monai/deploy/operators/monai_seg_inference_operator.py
+++ b/monai/deploy/operators/monai_seg_inference_operator.py
@@ -14,16 +14,15 @@ import logging
 import os
 from pathlib import Path
 from threading import Lock
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
+import torch
 
 from monai.deploy.utils.importutil import optional_import
 from monai.utils import StrEnum  # Will use the built-in StrEnum when SDK requires Python 3.11.
-from monai.utils import BlendMode, PytorchPadMode
 
 MONAI_UTILS = "monai.utils"
-torch, _ = optional_import("torch", "1.5")
 np_str_obj_array_pattern, _ = optional_import("torch.utils.data._utils.collate", name="np_str_obj_array_pattern")
 Dataset, _ = optional_import("monai.data", name="Dataset")
 DataLoader, _ = optional_import("monai.data", name="DataLoader")
@@ -42,7 +41,7 @@ Compose_, _ = optional_import("monai.transforms", name="Compose")
 # Dynamic class is not handled so make it Any for now: https://github.com/python/mypy/issues/2477
 Compose: Any = Compose_
 
-from monai.deploy.core import AppContext, ConditionType, Fragment, Image, OperatorSpec
+from monai.deploy.core import AppContext, Condition, ConditionType, Fragment, Image, OperatorSpec, Resource
 
 from .inference_operator import InferenceOperator
 
@@ -54,11 +53,6 @@ class InfererType(StrEnum):
 
     SIMPLE = "simple"
     SLIDING_WINDOW = "sliding_window"
-
-
-# define other StrEnum types
-BlendModeType = BlendMode
-PytorchPadModeType = PytorchPadMode
 
 
 # @md.env(pip_packages=["monai>=1.0.0", "torch>=1.10.2", "numpy>=1.21"])
@@ -90,9 +84,10 @@ class MonaiSegInferenceOperator(InferenceOperator):
     @staticmethod
     def filter_sw_kwargs(**kwargs) -> Dict[str, Any]:
         """
-        Returns a dictionary of named parameters of the sliding_window_inference function that are:
-            - Not explicitly defined in the __init__ of this class
-            - Not explicitly used when calling sliding_window_inference
+        Returns a dictionary of named parameters of the sliding_window_inference function that:
+            - Are not explicitly defined in the __init__ of this class
+            - Are not explicitly used when calling sliding_window_inference
+            - Can be successfully converted from Python --> Holoscan's C++ layer
 
         Args:
             **kwargs: extra arguments passed into __init__ beyond the explicitly defined args.
@@ -101,19 +96,30 @@ class MonaiSegInferenceOperator(InferenceOperator):
             filtered_params: A filtered dictionary of arguments to be passed to sliding_window_inference.
         """
 
+        logger = logging.getLogger(f"{__name__}.{MonaiSegInferenceOperator.__name__}")
+
         init_params = inspect.signature(MonaiSegInferenceOperator).parameters
 
         # inputs + predictor explicitly used when calling sliding_window_inference
-        explicit_used = ["inputs", "predictor"]
+        explicit_used = {"inputs", "predictor"}
+
+        # Holoscan convertible types (not exhaustive)
+        # This will be revisited when there is a better way to handle this.
+        allowed_types = (str, int, float, bool, bytes, list, tuple, torch.Tensor, Condition, Resource)
 
         filtered_params = {}
         for name, val in kwargs.items():
+            # Drop explicitly defined kwargs
             if name in init_params or name in explicit_used:
-                # match log formatting
-                logger = logging.getLogger(f"{__name__}.{MonaiSegInferenceOperator.__name__}")
-                logger.warning(f"{name!r} is already explicity defined or used; ignoring input arg")
-            else:
-                filtered_params[name] = val
+                logger.warning(f"{name!r} is already explicitly defined or used; dropping kwarg.")
+                continue
+            # Drop kwargs that can't be converted by Holoscan
+            if not isinstance(val, allowed_types):
+                logger.warning(
+                    f"{name!r} type of {type(val).__name__!r} is a non-convertible kwarg for Holoscan; dropping kwarg."
+                )
+                continue
+            filtered_params[name] = val
         return filtered_params
 
     def __init__(
@@ -125,10 +131,11 @@ class MonaiSegInferenceOperator(InferenceOperator):
         post_transforms: Compose,
         app_context: AppContext,
         model_name: Optional[str] = "",
-        overlap: float = 0.25,
         sw_batch_size: int = 4,
-        mode: Union[BlendModeType, str] = BlendModeType.CONSTANT,
-        padding_mode: Union[PytorchPadModeType, str] = PytorchPadModeType.CONSTANT,
+        overlap: Union[Sequence[float], float] = 0.25,
+        sw_device: Optional[Union[torch.device, str]] = None,
+        device: Optional[Union[torch.device, str]] = None,
+        process_fn: Optional[Callable] = None,
         inferer: Union[InfererType, str] = InfererType.SLIDING_WINDOW,
         model_path: Path = MODEL_LOCAL_PATH,
         **kwargs,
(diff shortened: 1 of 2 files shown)


monai/deploy/operators/monai_seg_inference_operator.py
```diff
@@ -47,13 +47,29 @@
 __all__ = ["MonaiSegInferenceOperator", "InfererType", "InMemImageReader"]
 
 
+class BlendModeType(StrEnum):
```
MMelQin, line 50: Can we simply use the enum from MONAI core instead of redefining it and needing to handle the mapping (just in case the core decides to update the str enum)? Also applicable to PytorchPadModeType. I think it is fine to expose some of the innards of this class, and further add in the inline doc that internally it uses the MONAI inference utils functions for sliding window and simple inference, and the params need to be as expected by them.

Similar to the conditional import of the sliding_window_inference, we can do
```
BlendModeType, _ = optional_import("monai.utils.BlendModeType", name="BlendModeType")
PytorchPadModeType, _ = optional_import("monai.utils.BlendModeType", name="PytorchPadModeType")
```

monai/deploy/operators/monai_seg_inference_operator.py
```diff
@@ -320,6 +367,8 @@ def compute_impl(self, input_image, context):
                         roi_size=self._roi_size,
                         sw_batch_size=self.sw_batch_size,
                         overlap=self.overlap,
+                        mode=self._mode,
+                        padding_mode=self._padding_mode,
```
MMelQin, line 371: It is good to add these couple params as explicitly supported named args to the MONAI inference function. In addition, I had the idea of extending the support to all other params on the MONAI inference functions by passing the (filtered) **kwargs down to the functions. I will provide a static function for the filtering in the general comment section.

bluna301:
@MMelQin many thanks for your review! Completely agree with all of your suggestions, and appreciate the base filtering function you provided for handling the implicit parameters. I will work on integrating your suggestions into the operator.

monai/deploy/operators/monai_seg_inference_operator.py
bluna301: I initially had `BlendMode` and `PytorchPadMode` as optional imports from MONAI utils like you suggested, but during type checking, `mypy` was struggling to deal with type checking of optional imports. I ended up having these as required imports, but these do now rely on the MONAI utils definition.

```python
from monai.utils import BlendMode, PytorchPadMode

.....

# define other StrEnum types
BlendModeType = BlendMode
PytorchPadModeType = PytorchPadMode
```

Happy to change these back to optional imports if you like, just let me know.

monai/deploy/operators/monai_seg_inference_operator.py
bluna301: Filtering functionality added. Tested this out with a few example cases, and the desired behavior was displayed. Here is an example - we can see `input` and `predictor` inputs are ignored, and that `buffer_steps` and `buffer_dim` parameters are successfully passed to `sliding_window_inference` to produce a `ValueError` (values chosen purposefully to produce the error):

```python
# delegates inference and saving output to the built-in operator
# parameters pulled from inference.yaml file of the MONAI bundle
infer_operator = MonaiSegInferenceOperator(
    self.fragment,
    roi_size=(96, 96, 96),
    pre_transforms=pre_transforms,
    post_transforms=post_transforms,
    overlap=0.25,
    app_context=self.app_context,
    model_name="",
    inferer=InfererType.SLIDING_WINDOW,
    sw_batch_size=1,
    mode=BlendModeType.GAUSSIAN,
    padding_mode=PytorchPadModeType.REPLICATE,
    model_path=self.model_path,
    inputs="test",
    predictor="testing",
    buffer_steps=1,
    buffer_dim=4
)
```

```python
2025-08-01 19:04:15,548] [INFO] (ct_totalseg_operator.CTTotalSegOperator) - TorchScript model detected
[2025-08-01 19:04:15,548] [WARNING] (monai_seg_inference_operator.MonaiSegInferenceOperator) - 'inputs' is already explicity defined or used; ignoring input arg
[2025-08-01 19:04:15,548] [WARNING] (monai_seg_inference_operator.MonaiSegInferenceOperator) - 'predictor' is already explicity defined or used; ignoring input arg

....

[2025-08-01 19:04:18,125] [INFO] (monai_seg_inference_operator.MonaiSegInferenceOperator) - Input of <class 'monai.data.meta_tensor.MetaTensor'> shape: torch.Size([1, 1, 270, 270, 204])
[error] [gxf_wrapper.cpp:118] Exception occurred for operator: 'ct_totalseg_op' - ValueError: buffer_dim must be in [-3, 3], got 4.

At:
  /home/bluna301/miniconda3/envs/ct-totalsegmentator/lib/python3.9/site-packages/monai/inferers/utils.py(142): sliding_window_inference
  /home/bluna301/ct-totalsegmentator-map/my_app/monai_seg_inference_operator.py(435): compute_impl
  /home/bluna301/ct-totalsegmentator-map/my_app/ct_totalseg_operator.py(226): compute

[error] [entity_executor.cpp:596] Failed to tick codelet ct_totalseg_op in entity: ct_totalseg_op code: GXF_FAILURE
[warning] [greedy_scheduler.cpp:243] Error while executing entity 28 named 'ct_totalseg_op': GXF_FAILURE
[info] [greedy_scheduler.cpp:401] Scheduler finished.
[error] [program.cpp:580] wait failed. Deactivating...
[error] [runtime.cpp:1649] Graph wait failed with error: GXF_FAILURE
[warning] [gxf_executor.cpp:2428] GXF call GxfGraphWait(context) in line 2428 of file /workspace/holoscan-sdk/src/core/executors/gxf/gxf_executor.cpp failed with 'GXF_FAILURE' (1)
[info] [gxf_executor.cpp:2438] Graph execution finished.
[error] [gxf_executor.cpp:2446] Graph execution error: GXF_FAILURE
  File "/home/bluna301/miniconda3/envs/ct-totalsegmentator/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/bluna301/miniconda3/envs/ct-totalsegmentator/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/bluna301/ct-totalsegmentator-map/my_app/__main__.py", line 25, in <module>
    CTTotalSegmentatorApp().run()
  File "/home/bluna301/ct-totalsegmentator-map/my_app/app.py", line 61, in run
    super().run(*args, **kwargs)
  File "/home/bluna301/ct-totalsegmentator-map/my_app/ct_totalseg_operator.py", line 226, in compute
    seg_image = infer_operator.compute_impl(input_image, context)
  File "/home/bluna301/ct-totalsegmentator-map/my_app/monai_seg_inference_operator.py", line 391, in compute_impl
    d[self._pred_dataset_key] = sliding_window_inference(
  File "/home/bluna301/miniconda3/envs/ct-totalsegmentator/lib/python3.9/site-packages/monai/inferers/utils.py", line 142, in sliding_window_inference
    raise ValueError(f"buffer_dim must be in [{-num_spatial_dims}, {num_spatial_dims}], got {buffer_dim}.")
ValueError: buffer_dim must be in [-3, 3], got 4.
```

bluna301:
@MMelQin PR has been updated - please see comments for detailed updates.

I did have one remaining question about the `device` parameter - I see this is [hardcoded](https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/dec9305e772f6317f28cca8af7de5138c43d66b5/monai/deploy/operators/monai_seg_inference_operator.py#L307C9-L307C78) into the operator, and this is also a `sliding_window_inference` parameter. Is this something we need to address?

MMelQin (commented):
LGTM. Thanks for expanding the support.

bluna301:
@MMelQin - I updated the PR following your guidance, but made some additional tweaks. Admittedly there are some signficant changes in this commit - I default to you regarding implementing all of these changes.

I checked all argument types explicitly defined by `sliding_window_inference`; those that would fail Holoscan conversion, I defined as explicit parameters in the operator (these were `sw_device`, `device`, and `process_fn`). Now, this operator will accept all possible explicitly defined `sliding_window_inference` args, which was the initial intention of this PR. Through my testing, I saw that these types failed conversion:
- `None`
- `device` (`torch.device`)
- `Callable` (function)
- 3D (and higher) NumPy arrays
- `range()`

I added the Holoscan conversion filter to the previously defined `filter_sw_kwargs` method. Thus, this method now filters out any repeat/internally defined parameters, as well as filters out any arguments whose type would fail Holoscan conversion. I did find that lists/tuples with mixed types and nested list/tuples would fail conversion, but I think some onus should be on the user to enter expected SWI inputs/formats (we state as much in the comments). I left out 1D and 2D numpy arrays, even those these are accepted by Holoscan.

I noticed that some of the setter type checks (e.g. `overlap`, `sw_batch_size`) were not occurring in the operator during runtime due to how explicit parameters were defined (public/private delineations were used liberally). I updated parameters to be private (`_`) or public based on their behavior, and now type checking at runtime is occurring.

I updated the existing `overlap` and `inferer` checks; the former was not aligned with the validation checks present in `sliding_window_inference`, while the latter was missing the string type check.

Lastly, `torch` was changed from an optional to a required import - the existing operator has several places where this dependency is needed hard-stop.

MMelQin:
Thanks @bluna301. I'll be reviewing the changes. Also, I've created a feature enhancement on Holoscan SDK, to support a Python function to test if an arg can be converted.

monai/deploy/operators/monai_seg_inference_operator.py
```diff
@@ -74,6 +81,47 @@ class MonaiSegInferenceOperator(InferenceOperator):
     # For testing the app directly, the model should be at the following path.
     MODEL_LOCAL_PATH = Path(os.environ.get("HOLOSCAN_MODEL_PATH", Path.cwd() / "model/model.ts"))
 
+    @staticmethod
```
MMelQin, line 84: For this method, I think we need to filter out two dictionaries to avoid stripping kwargs unnecessarily, with these assumptions
- keyword args that are used by the sliding_window_inference() stay with the derived class. They are not at all affected by if Holoscan can convert them to args in C++, so we can have any Python type that the sliding_window_inference() requires
- the left over kwargs, not in the params of sliding_window_inference(), are passed down to the base class __init__ if they can be converted

I put it in the code below for easier explanation. Also, the 2nd returned dict is used for the base class. 

```
   @staticmethod
    def filter_sw_kwargs(**kwargs) -> Tuple[Dict[str, Any], Dict[str, Any]]:
        """
        Filers the keyword arguments into a tuple of two dictionaries.
        - a dictionary of named parameters to pass to the sliding_window_inference function that:
            - Are not explicitly defined in the __init__ of this class
            - Are not explicitly used when calling sliding_window_inference
        - a dictionary of named parameters to pass to the base class __init__ of this class that:
            - Are not used when calling sliding_window_inference
            - Can be successfully converted from Python --> Holoscan's C++ layer
        Args:
            **kwargs: extra arguments passed into __init__ beyond the explicitly defined args.
        Returns:
            filtered_params: A filtered dictionary of arguments to be passed to sliding_window_inference.
            filtered_base_init_params: A filtered dictionary of arguments to be passed to the base class __init__.
        """

        logger = logging.getLogger(f"{__name__}.{MonaiSegInferenceOperator.__name__}")

        init_params = inspect.signature(MonaiSegInferenceOperator).parameters
        s_w_i_params = inspect.signature(sliding_window_inference).parameters

        # inputs + predictor explicitly used when calling sliding_window_inference
        explicit_used = {"inputs", "predictor"}

        # Holoscan convertible types (not exhaustive)
        # This will be revisited when there is a better way to handle this.
        allowed_types = (str, int, float, bool, bytes, list, tuple, torch.Tensor, Condition, Resource)
        filtered_params = {}
        filtered_base_init_params = {}
        for name, val in kwargs.items():
            # Drop explicitly defined kwargs
            if name in init_params or name in explicit_used:
                logger.warning(f"{name!r} is already explicitly defined or used; dropping kwarg.")
                continue
            elif name in s_w_i_params:
                filtered_params[name] = val
                logger.debug(f"{name!r} used in sliding_window_inference; keeping kwarg for inference call.")
                continue
            # Drop kwargs that can't be converted by Holoscan
            elif not isinstance(val, allowed_types):
                logger.warning(
                    f"{name!r} type of {type(val).__name__!r} is a non-convertible kwarg for Holoscan; dropping kwarg."
                )
                continue
            else:
                filtered_base_init_params[name] = val
                logger.debug(f"{name!r} used can be converted by Holoscan; keeping kwarg for base init.")
                continue
        return filtered_params, filtered_base_init_params
```

MMelQin (approved):
The enhancements look good to me. Please see the proposed changes to split kwargs into inference use and for base operator __init__, which I believe is the simplest ways to separate concerns.

bluna301:
@MMelQin thanks for your review - I will work on implementing the changes.

If I understand correctly, the filtered SWI implicit params (`filtered_params`) would be passed as a `kwarg` input during the `sliding_window_inference` call, while `filtered_base_init_params` would be passed onto the base `__init__`, leaving the code looking something like this:

```python

# def __init__():
    self._filtered_params, self._filtered_base_init_params = self.filter_sw_kwargs(**kwargs) # Filter keyword args
    
    # Pass filtered base init params
    super().__init__(fragment, *args, **self._filtered_base_init_params)

# def compute_impl():

    d[self._pred_dataset_key] = sliding_window_inference(
                            **self._filtered_params,  # Additional sliding window arguments
                        )
```

If `self._filtered_params` will never be passed for Holoscan conversion, I would propose eliminating the three new explicit params (`sw_device`, `device`, and `process_fn`) I added in my last commit - these were originally made explicit to avoid Holoscan conversion, but if we won't be passing them at all, we can safely remove these from the operator. I would prefer this to not pollute the operator with additional explicit params, and would have the added bonus of not adding any explicit params in this PR at all, which would hopefully minimally change documentation and reduce user confusion.

Let me know if you disagree, but will plan on removing these three explicit params otherwise.

MMelQin:
> @MMelQin thanks for your review - I will work on implementing the changes.
> 
> If I understand correctly, the filtered SWI implicit params (`filtered_params`) would be passed as a `kwarg` input during the `sliding_window_inference` call, while `filtered_base_init_params` would be passed onto the base `__init__`, leaving the code looking something like this:
> 
> ```python
> # def __init__():
>     self._filtered_params, self._filtered_base_init_params = self.filter_sw_kwargs(**kwargs) # Filter keyword args
>     
>     # Pass filtered base init params
>     super().__init__(fragment, *args, **self._filtered_base_init_params)
> 
> # def compute_impl():
> 
>     d[self._pred_dataset_key] = sliding_window_inference(
>                             **self._filtered_params,  # Additional sliding window arguments
>                         )
> ```
> 
> If `self._filtered_params` will never be passed for Holoscan conversion, I would propose eliminating the three new explicit params (`sw_device`, `device`, and `process_fn`) I added in my last commit - these were originally made explicit to avoid Holoscan conversion, but if we won't be passing them at all, we can safely remove these from the operator. I would prefer this to not pollute the operator with additional explicit params, and would have the added bonus of not adding any explicit params in this PR at all, which would hopefully minimally change documentation and reduce user confusion.
> 
> Let me know if you disagree, but will plan on removing these three explicit params otherwise.

Yes, I agree with your assessment.

monai/deploy/operators/monai_seg_inference_operator.py
bluna301: Added.

MMelQin (approved):
Thanks for the refinement to make it the best it can be.
I have also tested the changes in existing examples successfully. Ready to merge.

PR #554, opened 2025-09-19
Making MONAI Deploy app and MAP run as REST services on Aidoc AiOS

MMelQin:
Adds a lightweight Flask-based REST wrapper to run a MONAI Deploy app (AISpleenSeg) with status and callback support, plus scripts and docs to test locally.
- Introduces REST endpoints (/status, /process) that run the app in a background thread and optionally POST results to a callback URL
-  Provides a sample AISpleenSeg app wired into the wrapper, including pipeline composition and result reporting
- Adds local test script and documentation for setup, data download, and endpoint testing
- Package the REST service app as MAP with monai-deploy CLI and run the MAP container with `docker run` command for needing additional options such as volume mappings and environment variable passing.

### DIFF
--- a/platforms/aidoc/restful_app/app.py
+++ b/platforms/aidoc/restful_app/app.py
@@ -23,12 +23,14 @@ import requests
 from flask import Flask, jsonify, request
 from flask_wtf.csrf import CSRFProtect
 
+
 # Processing status enum per Aidoc API specification.
 class ProcessingStatus(str, Enum):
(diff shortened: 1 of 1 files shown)


MMelQin:
SonarQubeCloud / SonarCloud Code Analysis Quality Gate failure can be ignored, because the example in this PR deliberately reuses the same example code to illustrate how to make it an REST service for Aidoc.

PR #577, opened 2026-01-30
Resolve MAP runtime failure due to CUDA runtime lib missing by specifying cuda version when packaging MAP

MMelQin:
The "monai-deploy package" command for creating the MAP container image has changed to use CUDA 13 base image. However
- MD App SDK still remains dependent on holoscan SDK of the CUDA 12 version, holoscan-cu12
- While apps that have dependency on torch will not have runtime issue due to torch installation will make cuda12 runtime shared lib available in the image, but apps such as the Simple Imaging App that do not have dependency on torch will encounter runtime error
- MD App SDK will change to depend on CUDA 13 in the next release

So, the changes:
- The simple change is to specify the CUDA version with the packager command option `--cuda 12` so that the CUDA 12 version base image will be used, though the whole output of running the Jupyter notebook was added to this PR
- Unrelated changes to fix formatting complaints
- Updated commandline option value in the documents


bluna301:
@MMelQin - will review in full before EOD, but wanted to quickly comment to say I concur with your assessment and too found that explicitly specifying `CUDA 12` in the MAP build command resolved the issue when using `monai-deploy-app-sdk==3.5.0`, `holoscan-cu12==3.10.0`, and `holoscan-cli==3.10.0`. Running these MAPs with the MAR gave the expected behavior. Had to dig into the [Holoscan CLI - Package command](https://docs.nvidia.com/holoscan/sdk-user-guide/cli/package.html#:~:text=The%20CUDA%20version,of:%2012%2C%2013.) documentation to see the default parameter of `CUDA 13`.

Thanks for taking a look into this.

bluna301 (approved):
LGTM.

chezhia:
I was able to build a test MAP with cuda flag set to 12. However, Holoscan version should be 3.10 for the flag to work (holoscan-cli>=3.10.0 and holoscan-cu12>=3.10.0). The current requirements.txt and setup.cfg do not state this explicitly, consider editing those before merge..

MMelQin:
> I was able to build a test MAP with cuda flag set to 12. However, Holoscan version should be 3.10 for the flag to work (holoscan-cli>=3.10.0 and holoscan-cu12>=3.10.0). The current requirements.txt and setup.cfg do not state this explicitly, consider editing those before merge..

Great point! I will add some text in the Jupyter notebook example. Another thing is that I was thinking to update the Docs at the next release of the SDK as Doc build and publishing is one of the steps in releasing a new version, AND in the next release we move to CUDA 13 eliminating the need to specify --cuda on the commandline. However, did not have time to discuss this move to CUDA 13 in the WG meeting today, and PyTorch latest still by default uses CUDA 12.8 runtime lib. so we may not want to move to CUDA 13 yet.

